PR #5, opened 2026-09-23
Keep the full URL when editing the address

yuxino:
On a page with a query string, press ⌘L and Return. The address field has dropped the query, so Search opens a different page. I reproduced this on a filtered GitHub Issues page: the filter disappeared.

Before (real Search app, annotated):

https://github.com/user-attachments/assets/60075934-ca9c-4ef2-a9e5-9525558c6dca

After (annotated):

https://github.com/user-attachments/assets/795de39a-e668-41e8-bd5f-e5ba71f01f80

The editable field now uses the full current URL. The short tab label is unchanged.

Checked in the macOS app: ⌘L then Return keeps the filter. `swift build` passes.


driceroland:
Thanks for catching this one. Merged: ⌘L then Return keeps the whole address now.

PR #9, opened 2026-09-23
Floating video shows the whole picture, and its page lands back in the tab

### LINKED ISSUE #8: Floating video is black on YouTube, and the tab is empty after it lands
**What I did**

1. Play a YouTube video.
2. Press ⇧⌘P, or switch to another tab so the video floats on its own.
3. Press ⇧⌘P again.

**Expected:** the video plays in the floating window. After step 3, the page is back in its tab.

**What happened:** the floating window is black. If the page was scrolled, the whole window is black; otherwise a black band covers the top. After step 3 the tab is empty, with no page and no "playing in the floating window" line. It stays empty until the window is resized or another tab is picked.

**macOS:** 27.0 (26A428), Search 1.0

**Cause** (reproduced on `main` @ cc8aa58 in a `SEARCH_PROBE` build)

1. **Black window.** `Isolate.on` pins the video with `position: fixed`, but WebKit still clips it to its ancestors' overflow: `#movie_player { overflow: hidden }` and `#full-bleed-container { overflow-x: clip }`. In the 440×247 float, YouTube lays the player out at y=56, so the top 56pt is black. With the page scrolled, the player box is entirely off screen, so the whole window is black. The video's `getBoundingClientRect()` reads 0, 0, 440, 247 the whole time. Layout is correct; only the paint is clipped.
2. **Empty tab.** `Page` passes the same `tab.web` to `WebStage` before, during and after the float. SwiftUI sees no change, so it never calls `updateNSView` when the float lands. `Float.drop()` has already removed the web view from the panel, which leaves the view with no superview and no window. The stage's layout-time self-heal only runs when something else triggers a layout. Picking another tab works because `activeID` changes. Pressing ⇧⌘P again on the same tab, or the float's ×, does not.

I've opened a PR with a fix.

Chinteyley:
Fixes #8.

- **Black window** (`Float.swift`): while a video floats, every ancestor of the marked video gets `overflow: visible`, via `html.office-floating body :has([data-office-float])`. The rule sits in the same sheet, so `Isolate.off` removes it. It also follows the mark when the watch re-marks a new video. `html` and `body` keep `overflow: hidden`.
- **Empty tab** (`Stage.swift`): `Page` hands `WebStage` nil while `tab.floating` is true. Landing then changes nil to the page, which SwiftUI passes on to `updateNSView`, and the stage takes the page back right away.

Tested in a `SEARCH_PROBE` build on macOS 27.0 with a YouTube video:

| Case | Before | After |
|---|---|---|
| ⇧⌘P, page at top | 56pt black band | full picture |
| ⇧⌘P, page scrolled 900px | all black | full picture |
| Switch tabs (auto float), then pick the tab again | band; page back | full picture; page back, still playing |
| ⇧⌘P again on the same tab | empty tab | page back, still playing |

`swift build` passes, and the two changed files add no warnings. I only tested YouTube.


driceroland:
Thank you for the careful diagnosis and the fix. Merged, and it closes #8. It ships in the next update.

PR #16, opened 2026-09-23
Keep the address field in place while suggestions show

fschrhunt:
On a new tab, typing pushed the address field up. The field and the suggestion list were stacked in one column, so a list that grew re-centred the column and carried the field with it. The list now sits below the field instead of in that column, so it can arrive, change and leave without the field ever moving. The list's own animation is unchanged.

Preview: 

https://github.com/user-attachments/assets/e4359b95-99dd-4fa0-92d3-13d486b6e734

Checked on a new tab and over a page (⌘L): the list arrives as before, and the field stays put. swift build passes.


driceroland:
Thanks, this is much cleaner than stacking. Tested: the field holds exactly still with the list in and out. Merged. One follow-up on main: the container's `.animation(Motion.settle, value: browser.offers)` was what animated the list's arrival and its rows, since nothing sets `offers` inside an animation, so I put it back on the field.

PR #19, opened 2026-09-23
Keep text editing shortcuts from navigating away

yuxino:
On GitHub's sign-in page, type in the username field and press ⌘⇧← to select the text. Search goes back to the previous page instead. ⌘←/⌘→ also navigate away while editing text.

Before (real Search app):

https://github.com/user-attachments/assets/e879fd4c-aee6-477f-a872-ecca53a43bcb

After:

https://github.com/user-attachments/assets/9a9cc4c0-77eb-44ba-ad9c-a9da8c8fd339

These keys now stay with text fields, including the address field. Unshifted ⌘←/⌘→ still go back and forward when not editing.

Checked in the macOS app: text selection, moving to either end of a field, and back/forward navigation. `swift build` passes with no new warnings.


driceroland:
Thanks! Checked in a probe on Wikipedia's search field: ⌘⇧← selects the text and stays on the page, ⌘← in the field stays too, and ⌘← with nothing focused still goes back. Merged for 1.0.1.

PR #22, opened 2026-09-23
Follow the Mac's appearance unless the person picks one

mikuteto-dev:
New installs defaulted to Light, so on a Mac set to dark the window and every page's prefers-color-scheme stayed light until the person found Settings › Appearance. The default is now System: the browser follows the Mac, chrome and pages together, and an explicit choice still wins.


SamarthaB10:
> New installs defaulted to Light, so on a Mac set to dark the window and every page's prefers-color-scheme stayed light until the person found Settings › Appearance. The default is now System: the browser follows the Mac, chrome and pages together, and an explicit choice still wins.

I like this PR ngl

driceroland:
Thanks! Merged for 1.0.1 (I brought the branch up to date with main — only the changelog line conflicted). Checked in a probe on a Mac set to dark: a fresh install now follows it, and an explicit choice in Settings › Appearance still wins. This also lands together with #38, which moved the line right under yours to `NSApplication.shared`.

PR #24, opened 2026-09-23
A tab switcher with previews on ⌃Tab, off until turned on

farchanrifai:
Takes **A tab switcher with previews (⌃Tab held down)** off the roadmap.

## What it does

With **Settings › Tabs › Recently used tab switcher** on, holding ⌃ and pressing Tab shows up to ten tabs as pictures, in the order you last looked at them — the way Firefox and Arc do it. Letting go of ⌃ goes to the one picked.

- Tab / ⇧Tab and the arrow keys move through the tabs while ⌃ is held.
- Esc, or leaving the window, stays on the tab you're on.
- The pictures are the ones tab sleeping already takes, so nothing extra is drawn for this.

## Off by default

Per CONTRIBUTING, it's off until someone turns it on in Settings › Tabs. While it's off, ⌃Tab and ⌃⇧Tab step through the row exactly as they do on `main` today.

## Code

- `TabSwitcher.swift`: the recently-used order and the walk through it, with tests in `Tests/SearchTests/TabSwitcherTests.swift` (`swift test`; `Package.swift` gains the test target).
- `App.swift`, `Browser.swift`, `Tab.swift`, `Sleep.swift`: feed it, and draw the overlay.
- CHANGELOG gets its line under Unreleased, and ROADMAP loses the item.

## Tested

- Built on current `main`; `swift test` passes.
- Switcher off: ⌃Tab and ⌃⇧Tab step both ways through the row, checked with `./bench press` on a test run.
- Switcher on: the previews, moving through them, switching on release, and Esc were tried by hand.


farchanrifai:
https://github.com/user-attachments/assets/640b9673-4a18-4141-a287-df3228ea02f5

driceroland:
Thanks — careful work: off by default, tests, and the sleep snapshots reused rather than a new capture. It isn't going into 1.0.1, which is fixes and small things, and the keyboard handling it touches changed a lot tonight (Tab/⌃Tab, ⌘1–9 by key code for every layout, ⌃1–9 for Spaces), so it will need a rebase. We'll look at it properly after the release.

driceroland:
Thanks @farchanrifai — the ⌃Tab switcher is on the roadmap, and off by default is right. Two things before it can land: a rebase onto current main (the key handling changed), and, with the setting off, nothing should be recorded or published on tab changes (right now `previews` is reassigned on every switch even when it's off). Also note that ⌃Tab is used to step through tabs today, so the switcher should only take over when turned on. We'll skip the test target for now. Looking forward to it!

farchanrifai:
Thanks for the review! I've rebased this onto current main and made the changes you asked for:

- With the switch off, nothing is recorded or published on a tab change anymore. The order, the pictures and the `previews` updates only happen while it's on, and turning it off clears whatever was kept.
- ⌃Tab only takes over when the switcher is on. Off, it walks the row exactly as before.
- The test target is gone.

Let me know if anything else needs a tweak.

PR #25, opened 2026-09-23
Optional page-colored tab row

fschrhunt:
## Page-colored tab row

Adds an optional **Match the page color** setting for the horizontal tab row, off by default. The sidebar is unchanged. It keeps the toolbar-free window and the existing traffic-light placement.

The page reports its top-edge color from a Swift-embedded WebKit script. CSS colors update directly; animated headers pass their transition timing to the strip. Painted headers use a bounded snapshot fallback (at most eight sampled looks per page). The strip stays mounted across tab changes, so switching tabs changes the color at once; opening a blank tab no longer animates the page underneath its address field. The tint reads are throttled, snapshots wait for the page to settle, and the fallback waits were shortened after previewing them.

This incorporates the review on the original experiment: no toolbar, no `Bundle.module` script, and an opt-in setting. The sampling helper now lives with `PageTint.swift`. Merged current `main`, including folded tabs and the new bench controls, and resolved the four conflicts.

Preview: https://github.com/user-attachments/assets/0532bed7-671f-4527-a70f-c5b590fea8ac

Verified with `./build.sh release app` after the merge.


driceroland:
Thanks for asking before — here's the take on the direction. A bar that picks up the page's colour is appealing (Safari offers it), and if it comes, it should be an option, off by default. Two things would need to change:

- It brings back the window toolbar, which Search dropped on purpose: on macOS 26 a window with a toolbar gets almost twice the corner radius (31.5 pt against 17.5), which is why `Lights.swift` places the traffic lights by hand. A tint should work without it.
- Same `Bundle.module` issue as #29 — the script would need to be a Swift string.

If you'd like to pursue it as a small, optional tint without the toolbar, it's welcome — after 1.0.1.

fschrhunt:
@driceroland This is ready for review and merge now that 1.0.1 is out. I merged current `main` into the branch and resolved the tab-row, bench, settings, and WebKit-handler conflicts; the base commit is an ancestor of the PR head, and the 1.0.1 release app builds successfully.

Since the original experiment, the tint is opt-in and off by default, applies only to the horizontal tab row, keeps the toolbar-free window and existing traffic lights, and embeds the page script as a Swift string. CSS colors follow the page’s transitions; image/painted headers have a capped snapshot fallback. Tab switches update without replacing the whole strip, opening a blank tab no longer animates the page, and the fallback waits were tightened after previewing. The small snapshot sampler now lives in `PageTint.swift` rather than a separate `DominantColor.swift`. I updated the PR description to reflect the finished scope.

fschrhunt:
Here is a video showcasing the feature. Since the other expired. 


https://github.com/user-attachments/assets/41a4cf0c-cee3-44d9-bcc2-63fbb37e006d

PR #27, opened 2026-09-23
A middle-click on a tab closes it

lusqua:
Closes the roadmap item "Middle-click closes a tab."

The middle button on a tab closes it, as in other browsers — in the row across the top and in the column, rows and pinned squares alike. It goes through `Browser.close(_:)`, so a pinned tab is put down as with ⌘W, not removed.

SwiftUI has no middle-button gesture, so each tab gets a small NSView overlay (`MiddleClick`, in TabBar.swift) whose `hitTest` only answers for `otherMouseDown`/`otherMouseUp` with button 2 — the same idea as `DragStrip` refusing the tabs' run. Left click, drag-to-reorder and the right-click menu still reach the tab underneath. It closes on release, only if the pointer is still over the tab. A middle-click on a link in a page is unchanged (still opens a new tab).

Checked on a `SEARCH_PROBE` build, both layouts, with mouse events posted through the app's event queue: middle-click closed the tab under it; a pinned tab went to sleep and kept its place; press-and-release-elsewhere and button 3 closed nothing; left-click still selects, drags still reorder rows and pins, right-click still opens the tab menu. A hardware middle button also closed tabs in the row. `swift build` adds no warnings.


driceroland:
Thanks! Checked in a probe, both layouts: a middle-click closes the tab under it in the top row and in the column, the last tab leaves an empty one as ⌘W does, and a plain click still reaches the tab. Merged for 1.0.1 (I brought the branch up to date with main; only the changelog conflicted).

PR #29, opened 2026-09-23
Show link destinations on hover

fschrhunt:
I added a small bubble at the bottom of the page that shows where a link actually goes before you click it. It moves to the other corner if the pointer is down there, and it never blocks the page.

https://github.com/user-attachments/assets/ebd53482-eab0-4ad5-8ea7-d549fa2ce26a

Checked with `swift build`, `./build.sh release app`, and link hovers on GitHub, including long URLs and the corner swap.


driceroland:
Thanks — seeing where a link goes before clicking is worth having, and the corner swap is a nice touch. Two things before it can go in:

1. **Off by default**, with a switch in Settings. New features stay off until someone turns them on (see CONTRIBUTING) — Safari itself keeps its status bar hidden by default.
2. **`Bundle.module` won't find the script in the shipped app.** With `swift build`, SwiftPM's generated accessor looks for `Search_Search.bundle` at `Bundle.main.bundleURL` — the root of `Search.app`, not `Contents/Resources` (and a file at the root would break the signature) — and otherwise falls back to the absolute `.build` path of the Mac that built it. So it works on the build machine and traps at launch everywhere else. The rest of the app keeps its page scripts as Swift string literals (`Reader.script`, `ExtensionShims.script`…); doing the same here removes the resource and the `build.sh` change.

Happy to merge after that — not in 1.0.1, which is going out now.

fschrhunt:
Thanks for the review. I moved the script into `HoveredLink.script` in `StatusLine.swift` and removed the resource bundle and build-script changes.

On the opt-in point, I don't think this should need a switch. Seeing where a link goes before clicking feels like core browser behavior, especially when the link text doesn't reveal the destination. The bubble only appears while hovering a link, so it doesn't add persistent UI. I'd like to keep it on for everyone rather than put it behind a setting.

driceroland:
Thank you @fschrhunt! Two changes before merging: it's off by default (Settings › General › Show where links go), and when it's off no script is injected at all. I also removed the window-wide mouse-moved tracking; the bubble picks its corner once, when the hovered link changes. It still never takes clicks. Merging!

PR #30, opened 2026-09-23
The sidebar can hide by itself until the pointer reaches the edge

lusqua:
Closes the roadmap item "The sidebar hides by itself until the pointer reaches the edge, as an option on top of ⌘S."

A switch in Settings › Tabs, "Hide the sidebar until the pointer reaches the edge", shown while the tabs are down the left and off by default. With it on, the ⌘S fold is where the column rests: at launch, after switching layouts, and when the switch is flipped. The left edge brings it out over the page as before, and ⌘S still brings it out to stay and puts it away again.

Folded for good, the column waits for the pointer to rest on the edge for 0.15 s, so crossing the edge on the way to the Dock doesn't flash it. A ⌘S fold still comes out at once. The traffic lights now hide once the window exists, so a column folded at launch doesn't leave them over the page's corner (`hideLights()` returns early before there is a window). The column also stays out while a tab's address is being typed into its row — this last one applies to a plain ⌘S fold too.

Built on top of 1df70b5 (the lights riding on the column's spring, `Browser.peek(_:)`), not around it.

Checked on a test world, driving the pointer from inside the app and reading `bench probe` plus window captures and `bench film peek`: folded at launch with the lights hidden; out after resting on the edge, not when crossing it; back in after the pointer leaves; ⌘S out and back; kept out while a row is being edited; with the switch off, ⌘S and the edge behave as before. Not yet tried with a physical mouse. `swift build` adds no warnings. `bench ui hides on|off`, and `probe` reports it.


driceroland:
Thanks! Checked in a probe: with the switch on, the column rests folded and the traffic lights stay hidden, right after a relaunch too; they come out with the column at the edge and go with it; switched off, everything is back. It sits well on top of tonight's change that slides the lights in and out with the column. Merged for 1.0.1 (brought up to date with main: the changelog, and the bench's help line, which the Web Inspector change had just edited).

PR #32, opened 2026-09-23
A tab can be given a name of your own

theosementa:
Rename, in a tab's right-click menu and in Tabs › Rename Tab, types a name over the title in the tab itself.

It is the field that already edits a tab's address in place, told to mean something else for the length of one edit: the same pill, the same Return to keep and Escape to drop. Emptying it hands the tab back to whatever the page calls itself. No new field, no dialog — the whole change is one flag on the Browser, a `name` on the Tab, and a line in each of the two menus.

The name is the tab's, not the page's, so it survives being sent somewhere else, and it goes into the session beside the address and the pin — a tab you named is a tab you are keeping for a job. A pinned tab still shows its letter; its name is in the tooltip.

**Checked on a probe with `./bench`:** renaming through the menu writes the name into `session.json` at once, and a session holding a name comes back with it and is written out again unchanged. Rebased on `main` after Spaces landed and rebuilt: names sit beside `pin` in each space's own session file.


driceroland:
Thanks — reusing the address field for the name is exactly the kind of addition that stays out of the way. Checked in a probe that a named tab comes back with its name after a relaunch. One follow-up coming on main: Spaces, which landed after you opened this, rebuilds the rows of the spaces that aren't on screen from the session as well (`Browser.loadRow`), so the name gets carried there too. Merged for 1.0.1.

PR #34, opened 2026-09-23
The floating video window has no shadow

### LINKED ISSUE #33: Floating video window makes WindowServer do extra GPU work
When a video leaves its tab for the floating window, WindowServer does more GPU work for the same video than when it plays in the tab.

I measured the GPU time the kernel bills to WindowServer (coalition `gpu_time`, the number `powermetrics` reports) over 10 s windows. The test was Big Buck Bunny on YouTube in a test build of `main`, first with the tab in front, then after switching to another tab so the video floats.

| | 1080p | 4K |
|---|---|---|
| Playing in its tab | 22% | 19% |
| Floating, as it is now | 28% | 24% |
| Floating, `panel.hasShadow = false` | 16% | 16% |

The shadow is the cause. I tried an opaque window, square corners and a continuous corner curve, and none of them moved the number. Taking the shadow off is the only change that did. With no shadow, a floating video costs less than one in the tab, because the small window has fewer pixels to put on screen.

Search's own processes (the app plus its WebKit GPU, WebContent and Networking helpers) use the same CPU either way, about 15–25% depending on resolution. The 250 ms `setInterval` in `Isolate.on` doesn't show up in that number.

This came up because a user saw Activity Monitor's Energy Impact go from 40–100 to 200–300 when switching tabs with a video playing. On the same Mac I didn't reproduce a jump that big in Search's own processes. It showed up in WindowServer's share and nowhere else, and I don't know what accounts for the rest.

Measured on macOS 27, Apple Silicon, built-in Retina display.

AxxzyWasTaken:
Closes #33.

This is a one-line fix: `panel.hasShadow = false` in `Float.lift`.

A window with a shadow gets composited by WindowServer on every frame of the video. Without the shadow, a floating video costs less GPU time than the same video in its tab.

WindowServer GPU time (coalition `gpu_time`, 10 s windows, YouTube, switching to another tab so the video floats):

| | 1080p | 4K |
|---|---|---|
| In its tab | 22% | 19% |
| Floating, before | 28% | 24% |
| Floating, this PR | 16% | 16% |

I also tried an opaque window, square corners and `cornerCurve = .continuous`. None of them changed anything, so this PR leaves them alone. The rounded corners stay and so does the clear background. With no shadow the window's edge is less obvious against a dark background, but it's still clear against anything else.

Tested with `./build.sh release` and a `SEARCH_PROBE` world driven by `./bench`, on macOS 27 with the built-in Retina display. The build adds no new warnings.


driceroland:
Thanks — and for measuring it rather than guessing: 28% → 16% of WindowServer's GPU time is a real difference for a window that is often left playing for an hour. The edge is a bit softer on dark backgrounds; worth it. Merged for 1.0.1.

PR #36, opened 2026-09-23
Your own keyboard shortcuts, in Settings › Shortcuts

farchanrifai:
Takes **Your own keyboard shortcuts** off the roadmap.

Settings › Shortcuts

## What it does

**Settings › Shortcuts** lists every command Search has, with its key.

- **Find:** search by name or key, or filter to only the Changed or Disabled ones.
- **Change a key:** click the box and press the new one. If another command already has that key, it asks before replacing. macOS's own keys (⌘Q, ⌘C, ⌘V…) are refused, and so is a key without ⌘, ⌥ or ⌃.
- **Disable Shortcut** takes the key off, and the menu item shows none. **Reset to Default** brings back both the key and the setting below.
- **The menus** and the About page's list of keys always show the current keys.
- **When a website uses the same key** (the gear beside each command, or `⋯` to set all at once):
  - **Use Search First** is the default and how it works today: the page never sees the key.
  - **Use Website First**: the page gets the key, and Search acts only if the page doesn't use it.
  - **Prompt After Using on Website**: like website first, and the first time a site keeps the key, a bar asks which should win from then on.

## Nothing changes by default

Every command keeps the key it has today, and Search still wins every key, until someone changes it in Settings. Only changes are stored, so a default that changes later still reaches everyone who never touched that shortcut.

## How it works

- **`Shortcuts.swift`** is one list of commands. The menus are drawn from it and the key monitor dispatches from it, instead of each key being written twice in `App.swift`.
- **Website first:** the key monitor hands the key straight to the page's view, because the SwiftUI hosting view never passes ⌘-keys down to it. When the page doesn't use a key, WebKit sends back the same event; the monitor, or `PageView.keyDown` for keys without ⌘, recognises it and runs the command.
- **`ShortcutsPage.swift`** is the Settings page. Settings is now the same width (780) on every page, so the list and the selected command fit side by side.
- **Tests** are in `Tests/SearchTests/ShortcutsTests.swift`, with the test target added to `Package.swift`. If #24 is merged first, the two will conflict on that same one line.
- CHANGELOG gets its line under Unreleased, and ROADMAP loses the item.

## Tested

- `swift test`: key reading (including AZERTY's top row and ⌘+ / ⌘=), saving and reloading, disable, reset, taking a key from another command, and the routing rules.
- On a test run with `./bench press`:
  - the default keys work as before;
  - New Tab moved to ⌥⌘T, then disabled, then reset;
  - ⌘K on a page that uses it and one that doesn't, with Website First and with Prompt.
- The Settings page by hand: recording, replacing, refusing ⌘Q, the gear menu, reset, disable and the filters.


driceroland:
Thanks — impressive scope: conflicts, disabling, the site-or-Search setting, tests. It rewrites all of App.swift's key handling, which moved a lot tonight (⌘1–9 by key code so AZERTY works, Tab/⌃Tab, ⌃1–9 for Spaces, ⌥⌘I/J/C for the inspector), so it can't go into 1.0.1 and would need a careful rebase. It's on the roadmap; after the release, a smaller first step — only remapping the menu commands, say — would be much easier to land.

driceroland:
Thanks @farchanrifai — your own shortcuts are on the roadmap and we'd like them. This branch is quite far behind main now (the key handling and menus have changed: the Web Inspector keys ⌥⌘I/J/C, Spaces ⌃1–9, ⌘S folding the top bar…), so it would need a rebase onto current main. While you're at it, could you make it lighter? Keep rebinding, clearing and resetting a command, but leave out the per-command "website first" and "prompt" modes, keep Settings at its current width, and skip the test target for now. Search tries to stay as minimal as possible. Happy to review again once it's rebased.

farchanrifai:
Thanks for the feedback, that made a lot of sense! I've rebased onto current main and slimmed it right down (+476/−38, down from +1030/−196):

- **Kept:** rebinding, clearing and resetting a command, plus a small search field and a "press again" check before a key is taken from another command.
- **Removed:** the website-first and prompt modes, the wider Settings (it's the usual width now), and the test target.
- The existing key handling in `App.swift` is untouched unless someone actually changes a key, and the menus read their keys from the same place.
- Only menu commands can be remapped; ⌘1–9, ⌃1–9, Tab and Esc stay as they are.

Happy to trim further if you'd like it even smaller!

PR #38, opened 2026-09-23
Search opens on macOS 14 again

### LINKED ISSUE #37: Quits on launch on macOS 14: NSApp is nil in Preferences.init
**What I did:** downloaded Search.dmg from officecommun.com/search, moved Search.app to /Applications, opened it.

**Expected:** a window.

**What happened:** it quits right away, no window, no dialog. Same when running the binary directly: `/Applications/Search.app/Contents/MacOS/Search` exits with 133 (SIGTRAP), on every launch.

**macOS:** 14.3 (23D56), Apple Silicon. Search 1.0, build 202609231444. I think this is the same crash as the 14.8.3 one on the roadmap.

There is no `~/Library/Application Support/Search/crash.log` — it dies before anything gets written. The system crash report has `EXC_BREAKPOINT / SIGTRAP` on the main thread. Symbolicated with the dSYM from the v1.0 release (UUID `57483994-D48F-398A-B03E-5BA13D248EBD`, matches the installed binary):

```
0  Search     Preferences.init()          Prefs.swift
1  Search     Browser.init()              Browser.swift:34
2  Search     @objc Browser.init()
3  SwiftUI    … StateObject / AttributeGraph update
```

The trapping instruction is a `brk #1` right after `ldr x20, [_NSApp]; cbz x20`, so the implicitly unwrapped `NSApp` is nil. In source that's `Sources/Search/Prefs.swift:134`:

```swift
NSApp.appearance = chosen.appearance
```

On macOS 14 SwiftUI builds the App's `@StateObject` before it has created the NSApplication, so `NSApp` isn't there yet. Newer macOS seems to create the app earlier, which is probably why it doesn't reproduce on your machine. #22 changes the default look to System, but it still goes through `NSApp` here, so it doesn't change this.

Minimal repro on 14.3 (plain SwiftUI app, `@StateObject` whose `init` sets the app's appearance): `NSApp.appearance = …` exits with 133, `NSApplication.shared.appearance = …` runs. `shared` creates the app if it doesn't exist yet, so the look is still applied before the first window, as the comment above that line wants.

Opening a PR with that one-line change right after this.

serhiitroinin:
Fixes #37.

On macOS 14 Search quits on launch with SIGTRAP. `Preferences.init` sets the chosen look through `NSApp`, but SwiftUI builds the App's `@StateObject` (Browser, and the Preferences it holds) before it has created the application, so `NSApp` is still nil and the implicit unwrap traps.

The fix is `NSApplication.shared.appearance` instead of `NSApp.appearance`. `shared` creates the application if it doesn't exist yet, so the look is still set before the first window, not deferred, as the comment there wants. Everything after it in `Browser.init` then sees a real `NSApp` too.

Also in this PR: the line under **Unreleased › Fixed** in CHANGELOG.md, and the macOS 14.8.3 item taken off ROADMAP.md.

**Tested:** I couldn't run `swift build` here: this Mac has Xcode 15.4 (Swift 5.10), and the package needs Swift 6 plus the macOS 15.4 SDK for the web extension APIs. What I did check on macOS 14.3: a minimal SwiftUI app with the same shape (`@StateObject` whose `init` touches the app's appearance) exits with 133 using `NSApp` and runs using `NSApplication.shared`. It would be good if someone with the current toolchain runs `swift build` before merging.


driceroland:
Thank you — a textbook diagnosis: symbolicated with the release dSYM, the trapping instruction read, a minimal repro on 14.3, and a one-line fix that keeps the look applied before the first window. Checked here that nothing touches `NSApp` before `Preferences` in `Browser.init`, and that Search still launches and applies its look on macOS 26 (probe with this PR merged together with #22, #34 and #19). It goes out in 1.0.1 — it's the crash another person reported on X from 14.8.3.

PR #39, opened 2026-09-23
A column of more tabs than the window holds scrolls

lusqua:
With more tabs than fit, the column's stack overflowed at both ends: rows ran under the traffic lights at the top, and the foot - the bookmarks, the extensions, the space's dot - went off the bottom of the window with nothing to scroll them back. The loose tabs and the New tab row now scroll between the pins and the foot, and the tab you go to (⌘1-⌘9, ⇧⌘], a link opened beside) is scrolled into view.

While the tabs fit it is the plain stack it always was (ViewThatFits), so the space under them still drags the window.

Checked on a test world at 390x890 with ./bench hit: with 90 tabs the rows scroll from just under the lights' strip to just above the foot, and nothing is drawn under either; with 5 tabs the rows end where they did and the rest of the column drags the window, as before.


driceroland:
Thanks! Spaces landed while this was open and turned the column into one page per space, so I brought your change into that shape (the scroll now sits inside the live page, around the loose tabs and New tab, with the pager's swipe offset outside it) — same behaviour as your PR. Checked in a probe at 700×460: with 60 tabs the rows scroll between the lights' strip and the foot with nothing drawn under either, ⌘1 and ⌘9 bring the first and the last into view; with 5 tabs the space under them still drags the window. Merged for 1.0.1.

PR #40, opened 2026-09-23
Every space along the foot of the column

lusqua:
Follows up on Spaces (bdc784a).

In the column, the space on screen was one dot and every switch went through its menu. Now each space is a dot of its colour along the foot, the way Arc keeps them:

- a click on a dot goes to that space;
- the one on screen is lit, and a click on it opens the same menu as before (new, rename, colour, downloads folder, delete);
- hovering shows the space's name and its ⌃ number.

With a single space there's nothing to choose between, so it's the single dot as before. The row across the top keeps its dot unchanged. The dots are narrower than the lone dot (20 pt) so five or six still fit beside the bookmarks at the column's minimum width.

`swift build` adds no warnings.


driceroland:
Thank you for this — the same idea landed from the Spaces work while this was open, then went one step further at Drice's request: one icon for the space at the foot, turning over as you switch, with reordering in its menu (8085ee3). So I'm closing this one as superseded. And thanks for your other three tonight — #27, #30 and #39 are all merged for 1.0.1.

PR #41, opened 2026-09-23
Allow multiple empty tabs

### LINKED ISSUE #35: Allow more than one empty tab
## Steps to reproduce

1. Open Search on a new empty tab.
2. Without entering an address or opening a website, press ⌘T or choose New Tab.

## Expected

Search creates another empty tab. Each explicit New Tab action can create a new empty tab.

## Actual

Search keeps the existing empty tab and focuses its address field. Repeated New Tab actions do not add tabs.

## Cause

`Browser.newTab()` returns early when the active tab is blank. The comment says this was intended to prevent a held ⌘T from producing many empty tabs, but it also blocks deliberate requests.

## Environment

- macOS 26.6.2
- Reproduced three times with one empty tab open

## Acceptance criteria

- ⌘T and the New Tab button each create a new empty tab, even when the active tab is empty.
- The new tab becomes active and its address field is ready for input.

SamarthaB10:
## Summary

- Let every New Tab action create a tab, including when the current tab is empty.
- Remove the early return that reused an empty tab.
- Add the fix to the Unreleased changelog.

Closes #35.

## Checks

- `swift build` passes. It reports an existing deprecation warning in `ExtensionShims.swift`; this change adds no warning.
- In a separate test build, two consecutive ⌘T presses increased the empty-tab count from three to four to five. The new tab became active and its address field had focus after each press.
- The new-tab button and ⌘T both call `Browser.newTab()`.
- `git diff --check origin/main...HEAD` passes.
- `swift test` reports that the package has no test target.


driceroland:
Thanks! Merged for 1.0.1, with one follow-up on main: each ⌘T now makes a tab even beside an empty one, as asked in #35, but ⌘T *held down* makes only the first — holding the key, not pressing it twice, was what the early return was there for. Checked in a probe: two presses, two tabs; a held key, one.

driceroland:
Follow-up: on reflection we went one step further — ⌘T now never leaves two empty tabs: an empty tab already open elsewhere in the row is brought to the end and opened, with what was typed in it cleared (730e9a6). Thanks for pushing on this; it's cleaner than both the old behaviour and a row of blank tabs.

PR #42, opened 2026-09-23
The field leaves an extension's new tab page

meta-boy:
## What happened

Install any extension that replaces the new tab page with `chrome_url_overrides.newtab` and say Keep It. I used [Tabby Cat](https://chromewebstore.google.com/detail/tabby-cat/mefhakmgclhhfbdadeojlkbllmecialg). Type an address or a search into the field on that new tab and press Return. Nothing happens, and the tab stays on `chrome-extension://…/public/index.html`.

## Why

Search shows an extension's page in a view built from that extension's `webViewConfiguration`, and WebKit only lets that view load the extension's own pages. `replace(_:going:)` already swaps such a tab for a normal one. It runs from `decidePolicyFor`, though, and WebKit only asks that about navigations the page starts. When the field calls `submit()` → `tab.go(to:)`, the load never reaches the policy check and dies quietly.

## Fix

Each `Tab` now records which extension its view was built for (`home`, nil for the web), set by a new `Tab(for: url)` initializer. `go(to:)` compares that with the destination. When they differ it calls `onCross`, and the browser swaps the tab in place with `replace(_:going:)`.

The check lives in `go(to:)`, so no caller has to remember it. The field, suggestions, ⌘⇧V, bookmarks, history rows, links from other apps, a reopened closed tab and `chrome.tabs.update` all get it without changes.

## Tested

I ran a test world with `./bench --world tabby` and loaded Tabby Cat unpacked.

- With Tabby Cat as the new tab, I pressed ⌘T, ⌘L, typed `example.com` and pressed Return. Before, the tab stayed on `chrome-extension://local-…/public/index.html`. After, it loaded `https://example.com/`.
- I opened the extension's page in a bench tab with `ext-page` and sent it to `https://example.org` with `go`. The tab was swapped and loaded the site.

I only tried an address, not a search, but a search takes the same path through `submit()`. `swift build` shows no new warnings, and CHANGELOG has a line under Unreleased › Fixed.


driceroland:
Thanks — a real bug, and the right fix (the same path `replace(_:going:)` takes). Changes to extensions are on hold until 1.0.1 is out; this one is first in line after.

meta-boy:
@driceroland Rebased onto the latest main (bbc0c6e), so the conflicts are resolved and the branch is up to date. I built it and ran the new-tab check again, and it still works. Ready to go live whenever you are.

driceroland:
Thank you @meta-boy! I reworked it slightly on top of main: instead of remembering which extension a tab was built for, `go(to:)` decides from the page the tab currently shows, like the newer `ExtensionTab.loadURL` does, so a tab opened with `window.open` from an extension page is covered too. And `Browser.replace` now keeps private tabs private. Checked the field, `window.open` and a private round trip. Merging!

PR #48, opened 2026-09-23
Extension popups open at their size, and their button closes them

nicostassart:
Two things about extension popups, both seen with a pinned extension whose popup fills itself in from a script (most of them do).

**Opening.** The popover was shown as soon as the document was built. At that moment the popup had nothing in it yet, so the first measure was a sliver (25 × 76 for the one I was using), and the popover came in narrow and short, then stretched down, then across, over about half a second as `grow()` caught up. The host view controller's `preferredContentSize` was also left at zero, so even the remembered size never took. Now the popover stands unseen (window alpha 0, not animating) while the page is measured again every 40 ms. Once two measures agree and the page has loaded, it fades in at that size in 0.15 s. It is up the whole time, so the page runs as it did. Slow pages still get a cap: seen at 0.6 s after the first measure, or 0.8 s after opening if the document takes longer than that. After it is seen, everything is as before.

**Closing from its button.** A transient popover closes on the press, and a press on its own button is outside it. The button's action then arrives on the release and `press()` opened it straight back up, so the button could never close its popup the way it does in Chrome. `popoverWillClose` now notes a press that landed on the extension's own button, and `press()` lets that click only close. The note is dropped on the next press, or after a second, so a press dragged off the button doesn't swallow the next click.

**Bench.** `./bench ext-tap ID` (test runs only) clicks a pinned button the way a hand does: the press goes through the app's event queue, where the transient popover sees it, and then the action comes. The release can't reach a button in a window that isn't key, and a test run's window stays behind. That's how both bugs were reproduced and checked. Before the fix, the second tap left the popup open. The popover window's size over time went 53×105 → 53×586 → 366×586. After the fix, it is at alpha 0 until it holds 362×586, then fades in.

`swift build` adds no warnings.


driceroland:
Thank you @nicostassart! #137 landed the button-closes-the-popup part in the meantime, and we'd rather not delay every popup behind an invisible measuring phase, so I reduced this to the fix that matters most: `preferredContentSize` is set when the popover is created, so the remembered size takes from the start. Merging!

PR #49, opened 2026-09-23
Search with the engine you pick in Settings, Google by default (#43)

### LINKED ISSUE #43: Let me choose the search engine (currently Google only)
Search always sends to Google. 
I'd like to pick a different search engine, e.g. DuckDuckGo, Startpage, Ecosia, Kagi or Bing, or enter my own.

can you pls add custom Search Engines ?

karadoganyi:
Closes #43.

Settings › General gets a **Search with** menu: Google (still the default), DuckDuckGo, Bing, Ecosia, Startpage, Kagi, or **Custom**, an address with `%s` where the words go, the way Chrome and Firefox do it. Nobody who leaves it alone sees any change.

### What follows the choice
Everything that searched with Google before:
- the address field, and the name on its search row
- editing a tab's address
- paste and go (⇧⌘V)
- `chrome.search.query` for extensions

### How it works
- `Google.swift` becomes `Engine.swift`: the six templates plus a few pure functions, Foundation only.
- The choice is saved in `Preferences` as `search.engine` and `search.custom`.
- `Browser.searchURL(for:)` and `Browser.destination(for:)` are the one path from the setting to a URL.

### Custom addresses
- A custom address has to be http or https, have a host, and contain `%s` after the host. Anything else (a typo, `javascript:`, `file:`) isn't used: searches stay on Google, and Settings says why.
- The words are encoded once. An address copied with its own spaces, `ü` or `[ ]` in the query still gets the words exactly as typed.

### Privacy
Nothing new leaves the Mac. The query goes to the engine you picked instead of Google, and there are still no suggestion requests.

### Testing
- 53 checks on `Engine`, run with `swiftc` against the file on its own. They cover every preset, the round trip of `c++ & rust`, `a+b=c`, `100% sure`, Turkish and Japanese, templates with `%s` in the path, bad custom addresses and the Google fallback. I left them out of the PR because the repo has no test target. Happy to add one if you'd like.
- Each preset's address answers (Kagi asks you to sign in first, as expected).
- In the app: DuckDuckGo from the address field, the search row's name, a custom Brave address, and a bad custom address falling back to Google.

### Not in this PR
Right-clicking selected text still offers macOS's own "Search With Google", which never went through `Google.swift`. It could follow this setting in a small follow-up.


driceroland:
Thanks — tidy work, and the validation of custom addresses is exactly right. Checked in a probe: a search typed into a tab goes to Google by default, to DuckDuckGo once picked, to a custom Brave address with `c++ & rust 100%` encoded once; a `javascript:` template is refused and the search stays on Google. I merged main into the branch and pointed one new caller at `destination(for:)` (the click-away edit added tonight still called `Google.destination`). Merged for 1.0.1.

PR #50, opened 2026-09-23
Pin extensions to the toolbar from Settings

### LINKED ISSUE #44: pin extensions from settings page
I was able to install chrome extensions despite the official website stating extensions arent supported yet but theres no way to pin the extension so i cant use protonpass for example

merttopuz:
Closes #44

The puzzle-piece menu was the only way to pin an extension to the toolbar; Settings › Extensions now has the same Pin to Toolbar / Unpin action, on hover next to Reload / Options / Remove. It reuses `Extensions.setPinned(_:_:)`, the same call the puzzle-piece menu already makes.

| Before | After |
| --- | --- |
| old | new |

## Test plan

- Settings › Extensions, hover a row: Pin to Toolbar / Unpin shows and toggles the puzzle-piece menu's pin state
- Left click, Reload, Options and Remove on the same row are unaffected


driceroland:
Thanks — small and exactly where people looked for it. Merged for the next update.

PR #52, opened 2026-09-23
Give search engines their own Settings page and add Brave and Qwant

### LINKED ISSUE #43: Let me choose the search engine (currently Google only)

vivescene:
## Summary

- Move the search-engine choice from General to its own Search settings page.
- Style the control and opened list using Search's existing palette, with a clear selected state and without the oversized automatic focus ring.
- Add Brave Search and Qwant while retaining the engines and custom URL option introduced in #49. Google stays the default and first in the list.

This follows #49 and the now-closed #43; it does not change search routing or send suggestion requests.
CleanShot 2026-09-23 at 10  45 20@2x

## Verification

- `swift build` passes (only pre-existing warnings).
- Checked URL generation and query escaping for Google, Kagi, Brave Search, DuckDuckGo, and Qwant with a focused Swift harness.
- Checked the Search page, expanded list, selection, and custom URL field in an isolated macOS test build.


driceroland:
Thank you @vivescene! We kept the engine as a native picker in General (Settings stays as minimal as we can), so I reduced this to the part that matters: Brave Search and Qwant in the list. Checked both land on the right results page with an accented query. Merging!

PR #54, opened 2026-09-23
feat: add tab groups, space emojis and extension controls

K-NRS:
Tabs can now be grouped in both the sidebar and horizontal strip, with collapse, rename, emoji or native icons, and a new-tab action per group. Spaces also accept emoji markers. This extends the existing spaces implementation: per-space sessions, current icons, shared or separate sign-ins, download folders, Control-1–9 and sidebar swiping are preserved. Groups, markers and custom tab names survive space switches and restart. Pinning removes membership; ungrouping keeps tabs. Selection and resizing keep the selected tab visible.

Search runs extension-declared shortcuts but previously had no way to change or clear them. Settings → Extensions now offers a recorder, Clear and Reset for each command, with bindings restored after reload, disable/re-enable and browser restart. The recorder rejects unsupported keys and conflicts with browser/menu/space shortcuts or another enabled extension. Dispatch uses exact modifiers, preserves non-US keyboard characters, skips private tabs and reuses the existing popup path.

Offscreen dictionary lookups can fail because document discovery returns no contexts and iframe messages cannot reach the hidden parent page. This adds owned offscreen lifecycle management, filtered context discovery, readiness before creation completes, iframe messages and replies, and cleanup on reload or disable. Lokal can translate without a temporary tab in the strip. Uninterested extension pages no longer preempt slower replies after ten seconds. Ordinary inactive tabs retain their existing behavior; `runtime.getContexts` currently reports Search-managed offscreen documents only.

Float Video keeps centered videos inside the floating viewport and preserves Netflix's separate subtitle layer above the picture. Returning to the page restores its original styles; unrelated player controls stay hidden while floating. No dependencies added.

Validation on the final combined packaged release:
- Release build and ad-hoc signature verification passed.
- All 602 checks passed with production scheduling: 111 group/session checks, 37 native group UI checks, 110 extension-shortcut checks, 13 background-tab checks, 315 offscreen checks and 16 floating-video checks.
- Group tests cover legacy space metadata, named tabs, per-space persistence, collapse, marker validation, shared/isolated WebKit cookies and local storage, private/bench exclusion, reopen, replacement and canonical reordering.
- Native heading clicks and sheet controls were exercised in both layouts. Native scroll bounds and inspected screenshots verify first-tab pin updates, emoji/icon rendering, selected-tab reveal and a 640px dark tab strip. Pointer dragging itself was not automated.
- Regressions were reproduced before fixes for cross-group move rejection, unavailable-space ghosts blocking reopen, closing to the wrong displayed neighbor and selected tabs leaving view after resize.

Earlier feature-specific verification retained from this PR:
- Unchanged Lokal 1.5.9 returned 43 Tureng rows for an uncached lookup and 46 after reload, using one offscreen document without changing visible tabs or foreground selection.
- Native Settings layout, shortcut restart persistence and floating/returned video snapshots were checked separately. Netflix floating playback and Return were confirmed by the user; Netflix subtitle visual confirmation remains pending.
- Offscreen discovery/lifecycle/iframe failures, premature empty replies, shortcut interception, Turkish-key misidentification, space-shortcut conflicts and video/subtitle issues were reproduced before their fixes.


driceroland:
Thank you @K-NRS, there's a lot of work here. I'm closing it: it bundles four unrelated changes, tab groups are always on (Search keeps new features off by default, and with Spaces we've decided against tab groups for now), and it ships test-only code inside the app. The offscreen-document and Float Video fixes might be worth their own small PRs from current main; happy to look at those separately.

PR #55, opened 2026-09-23
Allow extensions on private tabs, off by default

### LINKED ISSUE #47: Extensions: No option to allow extensions on private mode, they just won't work.
Version: 1.0
MacOS 27.0

So Ublock Origin Lite specifically won't work on any private pages, it just states "not a website".

If i had to guess you just disable extensions on private mode with no option to change that.

merttopuz:
Closes #47

A private tab was left out of the WKWebViewConfiguration that carries the extension controller, and also filtered out of every tab list the extension system itself tracks (visibleTabs, activeAdapter, the tab-open/activate notifications, and a page's right-click menu items). An extension asking what tab it was on found none of it, and answered "not a website" instead of doing anything.

Settings › Extensions now has an "Allow on private tabs" switch, off by default so nothing already relying on the old behavior changes underneath it. Turning it on lets a private tab through the same checks as any other tab.

Allow on private tabs switch in Settings › Extensions

## Known limitation

The toolbar badge count (an extension's blocked-request tally, for one) doesn't update on a private tab even with the switch on, though the popup itself now shows the right site and data. This looks like a WebKit-level detail of the private tab's non-persistent data store rather than anything gated in this file, and needs more digging than this PR covers.

## Test plan

- Settings › Extensions › Allow on private tabs off (default): a private tab's extension popup still says "not a website", unchanged from before
- Switch on, open a new private tab, visit a real page: the popup now shows the site and its data like any other tab
- A normal tab is unaffected either way


driceroland:
Merged, thank you @merttopuz! One follow-up on top (7th line of the CHANGELOG credits you): a private tab opened *before* the switch is turned on has no controller in its page (it can't be added later), so extensions are now only shown private tabs that actually carry it.

merttopuz:
Nice, that makes sense. Thanks for the follow-up and for the credit! Thank you @driceroland

PR #56, opened 2026-09-23
An address bar above the page, in the sidebar's mode

### LINKED ISSUE #15: Add top search bar
Image

kndpt:
Closes #15.

image

image

image

image

image

An optional bar over the page when the tabs are in the sidebar, like Arc's developer mode: back, forward, reload, the site, the page's title. Off by default: Settings › Tabs › Address bar above the page.

- Click the site: connection security (a step in: the certificate), copy address, print, zoom.
- Click the title: the address field (⌘L).
- 40 pt, with the traffic lights on its line.

New code in `Bar.swift`; the rest are hooks. Bench: `bar PATH [card|security]`, `ui bar on|off`, `hit X Y click`.


lusqua:
LGTM!

PeterTheMango:
this is nice but also maybe the ability to customize where we want it, because for some users (like myself) i prefer it to be part of the sidebar rather than the address bar for the extra screen real estate. Maybe nice to make it a feature switch like we can choose between None | Address Bar | Sidebar

CleanShot 2026-09-23 at 6 50 54 PM@2x

kndpt:
Thanks! Good idea @PeterTheMango
I'd keep this PR to the bar only, and do the sidebar placement as a follow-up: the switch would become None | Address bar | Sidebar, reusing the same site card.

driceroland:
Thank you @kndpt! We decided against a bar over the page: Search is built around the column or the bar across the top, never both, so a second row doesn't fit. But Drice loved your site card, so it's now on main (66faf14) with your credit in the changelog: clicking the tab you're on to change its address brings the card up under it (connection → certificate, Copy Address, Print, Zoom), in both layouts, and Right-click › Site Information… opens it too. Your `SiteCard` code, palette colours and certificate panel are used almost as you wrote them. Thanks a lot for this.

PR #57, opened 2026-09-23
The reading fill in the tab you're on can be turned off in Settings › Tabs

karadoganyi:
From the roadmap's "Next" list: *a setting to turn off the reading-progress fill in the tab you are on* (asked on X).

Settings › Tabs gets **Show how far you've read**, a switch that's on by default, so nothing changes unless someone turns it off. Off, the tab you're on stays plain grey however far down the page you scroll, in the row across the top and in the column.

### How
- `Preferences.showsReading`, saved as `tabs.reading` (default on), next to `sleepsTabs`.
- The two places that draw the fill, `TabPill` in TabBar.swift and `SideRow` in Side.swift, draw it only when the switch is on. Both already watch `prefs`, so flipping it takes effect at once, with a fade.
- Only the drawing is switched off. The page's scroll messages keep coming, because the same script also carries the messages swipe navigation uses, and so the fill is back at the right width the moment it's turned on again.
- CHANGELOG line added, and the item taken off ROADMAP.md.

### Testing
- A long page scrolled halfway, in both layouts: the fill shows with the switch on, goes when it's turned off, and comes back at the same width.
- The switch saves: `tabs.reading` is written as soon as it's flipped.


driceroland:
Thanks! On by default, so nothing changes for anyone who doesn't look for it — exactly the shape we want for options. Merged for the next update.

PR #61, opened 2026-09-23
Mute a tab from its right-click menu

merttopuz:
Every other browser lets you silence a tab from its speaker icon without stopping the video - WKWebView has no call for that (it can pause media, but pausing stops it rather than muting it), so this does it in the page's own JavaScript instead: every `<audio>` and `<video>` element, present now or added later (tracked with a MutationObserver, for players like YouTube's that build their `<video>` element after the page loads), gets its `muted` property set. It runs in every frame, and is re-armed on each navigation, so a tab muted before it went somewhere else is still muted once it gets there.

"Mute Tab" / "Unmute Tab" is in a tab's right-click menu, in both the row across the top and the column down the side. A muted tab's speaker icon turns to a slash so the row still says which tab it is, same spot the playing icon already used.

Mute Tab in a tab's right-click menu

## Test plan

- Play a video, right-click the tab, Mute Tab: audio stops, video keeps playing
- The tab's icon turns to a muted speaker
- Unmute Tab brings the audio back
- Navigate the muted tab to a new page: it stays muted
- An unmuted tab is unaffected


PeterTheMango:
Would love to see this too, also an indicator so we know which tabs are currently playing audio. Possibly just needing to click this to toggle the mute to remove the need to right click if the user doesnt want to.

CleanShot 2026-09-23 at 6 48 56 PM@2x

merttopuz:
I think there’s already an indicator for tabs playing audio, but it doesn’t seem like you can click it to mute the tab. When you hover over it, the close icon shows up instead. Making the audio icon clickable to mute/unmute directly would be nice. @PeterTheMango 

image

PeterTheMango:
Yeah I saw that too, i think the placement could be better to be placed before the app icon or before the title of the tab, since its placed where the close button is which is confusing when trying to mute vs close. Thanks for pointing it out though @merttopuz 

CleanShot 2026-09-23 at 7 05 25 PM

peter-bf:
Nice addition! Could Search use WebKit’s page mute (WKWebView’s _setPageMuted: and _mediaMutedState methods) instead? A tab’s mute setting belongs to the browser UI. Page JavaScript shouldn’t see it through its media elements.

The current script has a few drawbacks:

  - It changes .muted and fires volumechange, so the site can detect the mute.
  - A site can set .muted back to false; the MutationObserver only catches newly added elements.
  - The live toggle runs in the main frame, leaving loaded iframes with their previous state.
  - It misses Web Audio (sounds generated with AudioContext can go straight to the audio output without an <audio> or
    <video> element to mute).
    
example:
```swift
static func set(_ muted: Bool, on web: WKWebView) -> Bool {
    let get = NSSelectorFromString("_mediaMutedState")
    let set = NSSelectorFromString("_setPageMuted:")
    guard let getter = web.method(for: get),
          let setter = web.method(for: set) else { return false }

    typealias Read = @convention(c) (AnyObject, Selector) -> UInt
    typealias Write = @convention(c) (AnyObject, Selector, UInt) -> Void
    let state = unsafeBitCast(getter, to: Read.self)(web, get)
    let write = unsafeBitCast(setter, to: Write.self)
    write(web, set, muted ? state | 1 : state & ~UInt(1))
    return true
}
```

WebKit’s page mute is closer to how Chromium mutes a tab. I tried it locally in Search and it stayed muted across navigation without changing video.muted 👍

These are private WebKit selectors (but better implementation), so the JavaScript approach could remain a fallback if they’re unavailable.

merttopuz:
Updated per the feedback above (@PeterTheMango, thanks) - the mute badge now sits beside the favicon instead of sharing the cross's spot, and is directly clickable: no need for the right-click menu anymore. Fixed-size box too, so swapping between the playing/muted icon doesn't nudge the title.

mute badge in front of the title, hover state shown

merttopuz:
@peter-bf Thank you, this was the right call. Switched to WebKit's own page mute (\`_mediaMutedState\` / \`_setPageMuted:\`), your example, asked for first the way \`inspector(_:on:)\` already asks before it reaches for a name outside the public framework - the JavaScript approach is now only a fallback for a WebKit that doesn't have it.

Since the page mute holds by itself across whatever a tab navigates to next, only \`toggleMute()\` and \`arm(hiding:)\` need to reach for it now; the fallback script still gets re-armed on every navigation, since the page itself has to be told each time.

driceroland:
Thank you @merttopuz! Drice asked for the speaker itself to be the control, so: the speaker a tab already shows while it plays sound is now a small button in both layouts (a faint circle on hover, like the ×), speaker.slash while muted, nothing for a silent tab. Mute Tab stays in the right-click menu. `toggleMute()` uses `built`, so muting a sleeping tab doesn't wake it, and the mute is re-applied when its page is rebuilt; I kept only the guarded `_setPageMuted:` path. Checked with a WebAudio test page: muted flag set, kept across navigation and through sleep. Merging! (I kept your WebKit page-mute approach from the latest commits and merged main into the branch.)

PR #69, opened 2026-09-23
Bookmarks in the column, above the tabs

### LINKED ISSUE #58: Bookmarks in the column, above the tabs
In the sidebar's mode, bookmarks only live behind the bookmark button at the foot of the column, in a popover. With a lot of them (folders included), I'd like them in the column itself, above the tabs, as Arc does:

Image

- A **Bookmarks** section between the pinned tabs and the tabs, rows the same size as a tab's. A site opens on a click, a folder opens in place.
- Bookmarks and tabs scroll together when the column runs out of room.
- **Off by default**: Settings › Tabs › "Bookmarks in the sidebar". An unconfigured browser looks as it does today; the row across the top doesn't change.
- The button at the foot stays (Add This Page, Manage Bookmarks…).

I'm happy to build it if that's welcome!

kndpt:
Closes #58.

https://github.com/user-attachments/assets/783fdbf0-854b-4471-81db-26e29ce2b45e

Bookmarks in the column, above the tabs, the way Arc keeps them. Off by default: Settings › Tabs › Bookmarks in the sidebar.

- A tab dragged onto the bookmarks becomes one and stays open under it; a bookmark dragged down among the tabs becomes a tab again.
- A bookmark is open (a dot, a cross under the pointer to shut it) or shut; a click opens it in its own tab or goes back to it. They come back shut after a relaunch.
- Drag to reorder and file into folders; folders open in place; the section folds under its heading.

New code in `Shelf.swift`; the rest are hooks, plus `Bookmarks.move(_:into:before:)`. Bench: `shelf [seed|open|close|drop|fold|unfold]`, `ui shelf on|off`, `drag X Y TOX TOY`.


kndpt:
@driceroland  Question before review: `Spaces.swift` says bookmarks are shared by every space, and this PR makes them each space's own (a space is a context, like its tabs and pins). Is that in line with the project? If not, I'll drop `2b34d62` and keep them shared.

driceroland:
Thank you @kndpt for this, it's carefully built. We've decided to keep the column as quiet as possible (Spaces already split tabs into sets), so I'm closing it. Also, favourites becoming per-Space for everyone would change the Bookmarks menu and ⇧⌘B for people who never turn the shelf on, and deleting a Space would delete its bookmarks. Thanks a lot for the work and for asking the right questions.

PR #70, opened 2026-09-23
Esc closes the history panel

### LINKED ISSUE #67: History modal: slow launch, scroll stuttering, and Escape key does not close
macOS version: macOS 27.0

What I did:
* Opened the history view using the keyboard shortcut / menu item.
* Scrolled through the list of entries.
* Pressed the Escape key.

What I expected:
* The modal to appear instantly without delay.
* Smooth scrolling down the list.
* The modal to dismiss immediately upon pressing Escape.

What happened instead:
* The modal takes a noticeable amount of time to open.
* Scrolling stutters and drops frames.
* Pressing Escape does nothing; the modal stays open.

merttopuz:
Closes #67

Settings, Bookmarks and Passwords all close on Esc, each checked in the app's key handler - History (`recalling`) was the one Plate-based panel missing from that chain. Its own close button already said "Done   esc" in its tooltip; Esc just never did it.

## Test plan

- Open History (⌘Y or File › Show History…), press Esc: it closes, same as Settings/Bookmarks/Passwords already do
- Esc still does everything else it did before (closing Find, dropping a suggestion, canceling a tab rename, and so on)


driceroland:
Thanks! The same four lines Settings, Bookmarks and Passwords already have — merged for the next update. (The slow opening and the scrolling reported in #67 are being looked at separately, so I'm leaving #67 open.)

PR #71, opened 2026-09-23
A middle-click on a link opens it in a new tab

### LINKED ISSUE #45: middle click to close tab and open link in new tab
Feels wrong not being able to close a tab by middle clicking it like i do on all the other browsers and middle click to open link in new tab would be great too

Dodothereal:
Closes #45.

A middle-click on a tab has closed it since #27, but a middle-click on a link did nothing at all: the page kept the click, and no tab was opened.

## Why the old code could not have worked

Next to the ⌘-click check in `webView(_:decidePolicyFor:)` there was an `action.buttonNumber == 2` test meant to catch a middle-click. Two things were wrong with it.

WebKit hands the browser a `WKNavigationAction` for a ⌘-click and **none at all** for a middle-click — the page gets the click and the delegate is never called — so the test never ran. And it was comparing against the wrong number: WebKit answers `buttonNumber` with a **mask**, not a button index. From WebKit's `WebEventFactory::toNSButtonNumber`:

```
Left   -> 1 << 0  = 1
Right  -> 1 << 1  = 2
Middle -> 1 << 2  = 4
```

So `== 2` is the *right* button; the middle one is 4. Confirmed against the running app: a ⌘-click reports `buttonNumber = 1`, which is the Left mask.

## What this does

The page can see the click even where the delegate can't, so the page is asked. A new `MiddleRelay` follows the existing `ImageRelay` / `StoreRelay` pattern: an injected script listens for `auxclick` on the middle button, walks up to the enclosing `<a href>`, and posts the address. The relay opens it beside the current tab — `Browser.open(url, foreground: false)` — and leaves you where you are, the same thing ⌘-click does.

Guards, so a page cannot open tabs by asking:

- the event must be a real click, not a synthesized one (a page dispatching its own `auxclick` gets nothing);
- it is listened for on the way up, so a click the page has already answered with `preventDefault` is left alone;
- the message must come from the main frame, and the address must be `http` or `https`.

The stale `buttonNumber` test is removed and the comment beside it now says why, so it does not get put back.

## Testing

Driven through `./bench` against a `--test` world, with a real middle-click posted at both buttons in the running app:

| | result |
|---|---|
| middle-click on a link | opens in a background tab, `activeUnchanged: true` |
| five page-dispatched `auxclick`s | open nothing |
| ⌘-click | still opens beside, foreground false |
| ⌘⇧-click | still opens in front, foreground true |

`swift build`: no new warnings.

## Notes

- Main frame only, deliberately: a middle-click on a link inside an ad iframe is that frame's business, and its link is not this tab's to open.
- No dependency, no network, nothing new leaves the Mac.
- CHANGELOG entry added under **Unreleased** as CONTRIBUTING asks.


lusqua:
LGTM!

driceroland:
Thank you @Dodothereal, and for spotting that `buttonNumber == 2` was the right button. I added a few things: a middle-button navigation action is cancelled in the policy (so a future WebKit can't cause a double navigation), the click that stops auto-scroll no longer opens a link under it, SVG links, image-map areas and links in open shadow roots work too, and a private tab's middle-click opens a private tab. Middle-click on a tab still closes it. Checked with real middle-button events. Merging!

PR #73, opened 2026-09-23
Search selected text in a new tab, with the engine from Settings (#66)

### LINKED ISSUE #66: Bug: “Search with Google” opens Safari instead of a new Search tab
When selecting text on a webpage and choosing “Search with Google” from the right-click context menu, the Google search opens in Safari instead of opening in a new tab within Search.

**Steps to reproduce**
1. Open a webpage in Search.
2. Select/highlight some text.
3. Right-click the selected text.
4. Click “Search with Google”.
5. Observe where the search opens.

**Expected behavior**
The selected text should be searched on Google and the results should open in a new tab within Search, without leaving the browser.

**Actual behavior**
The Google search is opened in Safari, the system's default browser.

**Likely area to investigate**
The context-menu action appears to be handing the generated Google search URL to macOS's default URL handler instead of routing it through Search's own tab/navigation system.

karadoganyi:
Fixes #66.

Right-clicking selected text and choosing "Search with Google" opened the search in the Mac's default browser (Safari for most people), and always said Google whatever was chosen in Settings › General › Search with.

### Why
The item is WebKit's own (identifier `WKMenuItemIdentifierSearchWeb`, action `forwardContextMenuAction:`). WebKit hands the text to macOS, which opens the default browser, so nothing in Search ever saw the click.

### What changes
- `PageView.willOpenMenu` finds that item by its identifier, renames it "Search with <engine>" from Settings (the host for a custom engine), and points it at the page view.
- The selection is read when the menu opens, in WebKit's isolated script world so the page can't change it. The click then opens the search in a new foreground tab with `searchURL(for:)` and `open(_:foreground:)`, as "Open Image in New Tab" does.
- From a private tab, the search opens in a private tab too. `open(_:foreground:atEnd:shy:)` gains a `shy` argument, defaulting to false, so nothing else changes.
- Whitespace-only selections do nothing.
- When the main frame can't see the selection (text selected inside an iframe, a PDF), the click goes back to WebKit's own action, so those cases behave exactly as before.
- It works on macOS 14: the callbacks run Browser → Tab → PageView like `onImageMenu` and `onPull`, with no dependence on `Extensions.shared`.
- If a future WebKit renames the identifier, the item is left alone and behaves as it does today.

### Testing
- In an isolated test profile set to DuckDuckGo, the item reads "Search with DuckDuckGo" and opens a new tab in Search at duckduckgo.com. After switching to Google, the title and the search follow.
- From a private tab: the search opens in a private tab, and the query appears in neither History nor the saved session. A normal-tab search is recorded, as expected.
- Text selected in the page first, then inside an iframe: the page's old selection isn't searched, and WebKit's own action takes over.


driceroland:
Thank you @karadoganyi! I merged main (which now has `open(…, from:)` for private tabs) and open the search `from: tab`, so it stays private from a private tab. I also carried over an idea from #120: the selection is read from a focused text field or a same-site frame too (never a password field). Merging!

Issue #68, opened 2026-09-23
Tab folders in the sidebar


armondschneider:
## What

Add optional, named folders to the sidebar for grouping open tabs. From the sidebar, right-click to create a folder and give it a name; folders appear above the regular tab list. Drag tabs into a folder to group them.

## Why

When many tabs are open, a small number of named groups (for example, Work or Reading) would make the sidebar easier to scan without changing the default experience for people who do not want groups.

## First version

- Sidebar only; folders are created from its right-click menu.
- Folders hold open tabs, not bookmarks.
- Drag existing tabs into a folder.
- The feature is optional and off by default, with a Settings switch.

I would keep the first version focused on creating, naming, collapsing, and moving tabs into folders. Please let me know if this direction fits before I start implementation.

PR #76, opened 2026-09-23
Add optional tab folders to the sidebar

armondschneider:
## Summary
- Add an off-by-default Tab folders setting for the sidebar.
- Create and rename folders inline below pinned tabs; drag tabs into folders, with hover/drop feedback and collapsible headings.
- Keep the active tab visible when its folder is collapsed and save folder state with each space’s session.

## Testing
- `swift build` succeeds (existing warnings in unrelated files).
- Local macOS app exercised while refining the folder interaction; draft for further testing and review.

Refs #68


driceroland:
Thank you @armondschneider! Like the other tab-group PRs, I'm closing this one: with Spaces for separate sets of tabs, we'd like the column to stay as quiet as possible, and folders also replace the smooth reordering with a drag that can drop a tab's id into other places. Thanks for the idea and the care.

PR #78, opened 2026-09-23
A native Share menu for the page

### LINKED ISSUE #77: A native Share menu for the page
Safari has a Share button (Mail, Messages, AirDrop, Notes, whatever else is registered) — Search has nothing like it. No toolbar means no button for it there, but it could live in the File menu (Share…), and optionally as a door beside the bookmarks for anyone who wants it findable without the menu.

merttopuz:
Closes #77

Safari has a Share button — Mail, Messages, AirDrop, Notes, whatever else the Mac has registered. Search has no toolbar to put one on, so:

- **File › Share…** hands the front tab's address to `NSSharingServicePicker`. Always there, like Print.
- **Settings › Tabs › Share button**, off by default, puts a door for it beside the bookmarks, in both layouts. From there the picker opens right under the door, the same way the bookmarks dropdown does; from the menu, with no door to open under, it centers on the window instead.

## Test plan

- File › Share…: the native picker opens, centered on the window, with the front page's URL
- Settings › Tabs: Share button is off by default; turning it on adds a door beside the bookmarks in both the top row and the sidebar
- Clicking that door opens the same picker, right under the door
- Every other door in that row (back, forward, reload, bookmarks) still lines up evenly with it added


driceroland:
Thank you @merttopuz! Drice wanted to keep Search's tab row exactly as it is, so I reduced this to the menu entry: File › Share… opens macOS's share sheet for the page, from the top-right of the page. The optional icon, its setting and the Door parameters are gone, and the window is found through `Links.window` so it also works in test worlds. Merging!

PR #79, opened 2026-09-23
Mouse back/forward side buttons navigate the page

reloadedhead:
A mouse's back/forward side buttons did nothing. Trackpad two-finger swipe already worked (a custom `Pull`, not WebKit's own gesture — see `allowsBackForwardNavigationGestures = false`), but the equivalent for a mouse was never wired up.

`PageView.otherMouseDown` now reads `event.buttonNumber` — 3 for back, 4 for forward, the X11 button order most mouse drivers use regardless of platform — and calls `goBack()`/`goForward()` when history allows it, falling through to `super` for every other button so nothing else on the mouse is affected.

Tested on real hardware: works cleanly with the stock driver. Third-party mouse utilities (Logitech Options+, SteerMouse, etc.) that remap the buttons to their own per-app keyboard shortcuts sit ahead of this and are unaffected either way — same as it works for Chrome and doesn't for Safari today.

PS: I loved this browser. The back/forward buttons not working was driving me crazy 😅.


amol909:
Tested this on real hardware: a Logitech G402 over USB, no mouse software installed, macOS 27.0 (26A428).

The side buttons arrive as `otherMouseDown` with `buttonNumber` 3 (back) and 4 (forward). With this PR applied to current `main`, both navigate the page, and a press with no history in that direction does nothing.

The `Tab.swift` change still applies cleanly to `main`. Only `CHANGELOG.md` conflicts, because **Unreleased** now has its own `### Added` section, so the line just needs to move under it.

Thanks for this. I was missing it too.

driceroland:
Merged, thank you! Checked together with today's other PRs in a combined build, in a hidden test world: it builds and runs; the side buttons themselves can't be pressed in a hidden test run.

PR #87, opened 2026-09-23
Frame rate setting: 60, 120, or 120 while plugged in

### LINKED ISSUE #75: Bug: Browser appears capped at 60 FPS

Reproduction:  
3. Run the FPS test.

AxxzyWasTaken:
Closes #75.

I noticed Search was stuck at 60fps on my 120 Hz MacBook Pro. It turned out to be WebKit's "Prefer Page Rendering Updates near 60fps" feature flag, which Safari also ships with on. I added Settings › General › Frame rate with three choices:

- **60**: the default, same as now.
- **120**: turns the flag off for every page.
- **120 plugged in**: 120 on the power adapter, 60 on battery. I listen for macOS's power source notification, so nothing polls.

I kept it off by default, per CONTRIBUTING.

## What it costs

I ran a short test on my MacBook Pro with the window at 1728×994 and Search in front. These are Activity Monitor energy impact numbers for Search's own processes:

| Page | 60 | 120 |
|---|---|---|
| Static page | 0.7 | 2.1 |
| One CSS animation | 9–11 (60fps) | 15 (120fps) |
| 1080p video + rAF counter | 20 (60fps) | 23 (120fps) |

WindowServer stayed about the same. A page that animates costs me roughly half again as much at 120, which is why I didn't make it the default and why I added the plugged-in option.

## One rough edge

WebKit reads the flag when a page is made, so new tabs and reloads get the new rate right away. Tabs I already had open didn't fully follow. Going up to 120, they caught up the next time I switched to them. Going back down to 60, they only changed after a reload. That applies to the plugged-in option too: if I unplug, my open tabs stay at 120 until I reload them.

I tried hiding and showing the view to make WebKit re-read it, but that only worked when the page actually left the screen. Reloading every tab would lose typed text, so I left it and mentioned it in the setting's description. If you know a cleaner way to get WebKit to re-read it on a live page, I'm happy to change it.

## Notes

- The flag isn't public API. I set it through `WKPreferences._features` / `_setEnabled:forFeature:` and check `responds(to:)` at each step, so a WebKit without it just stays at 60.
- I added `./bench ui framerate standard|smooth|plugged` so it can be set from a script.
- It builds with no new warnings. With bench and a rAF counter I checked that it's 60 by default, 120 after choosing 120, that the choice survives a relaunch, and that "120 plugged in" runs at 120 on AC. I haven't tested the battery side yet.
- I added a CHANGELOG line under Unreleased.


driceroland:
Thank you @AxxzyWasTaken, and for measuring the battery cost. We kept 60 fps as the default like Safari, but reduced it to one switch: Settings › General › Pages at 120 Hz, off by default. When it's off the private flag is never touched; the feature lookup is cached (thanks to #94 for that idea) and `./bench probe` reports `prefersNear60FPS`. The power-source mode is gone, to keep Settings minimal. Checked in a test world: untouched when off, set on new and open pages when on, reset when turned off again. Merging!

PR #89, opened 2026-09-23
The bookmarks popover holds still while you reach for a bookmark

### LINKED ISSUE #88: Bug: Bookmarks popover closes when hovering over a bookmark
Clicking the bookmark door at the bottom of the column opens the bookmarks popover as expected, but moving the pointer onto the bookmarks inside it closes the popover. That makes it hard to reach any bookmark, or "Add This Page" / "Manage Bookmarks…" buttons at the bottom of the popover.

Reproduction:
  1. Click the bookmark icon next to the home button at the bottom of the tab column.
  2. The bookmarks popover opens.
  3. Move the pointer up over the list of bookmarks.
  4. The popover closes before anything can be clicked.

merttopuz:
## Summary
- The bookmarks popover (`BookmarksDropdown` in `Sources/Search/Bookmarks.swift`) closed as soon as the pointer moved over the bookmark list, making it impossible to click a bookmark or the "Add This Page" / "Manage Bookmarks..." buttons.
- The popover's `ScrollView` was missing `.fixedSize(horizontal: false, vertical: true)` after `.frame(maxHeight: 360)`. Without it, each row's per-item hover animation caused SwiftUI to re-measure the scroll view's content size, which AppKit's `NSPopover` picked up as a size change and dismissed the popover on.
- Fixed by pinning the content size with `.fixedSize(horizontal: false, vertical: true)`, matching the identical pattern already used by the extensions popover (`ExtensionMenu` in `Sources/Search/Extensions.swift`), which does not have this bug.

Fixes #88

## Test plan
- [x] `swift build` succeeds
- [ ] Manually verified: open the bookmarks popover from the bottom of the tab column, hover over bookmarks, confirm the popover stays open and bookmarks/footer buttons are clickable


driceroland:
Thanks @merttopuz! Holding this one for a moment: the `.fixedSize` line is harmless (it matches the extensions menu), but a comment on #88 says the extensions menu closes the same way, and only when the sidebar hides itself. That points at Fold: moving into the popover (a window of its own) counts as leaving the column, so after the grace delay the column folds and takes the popover with it. #115 just reworked how the hidden sidebar follows the pointer, so I'd like to check the popover against that first. If you can reproduce #88 on current main, a note on which setup (hidden sidebar or not) would help a lot.

merttopuz:
Rebased onto main (so this now includes #115) and reproduced #88 on it: only with "Hide the sidebar until the pointer reaches the edge" on. With it off (sidebar always visible) it never happens.

Traced it with some added logging: `NSWindow.windowNumber(at:belowWindowWithWindowNumber:)` never resolves a popover to the topmost window at the pointer's screen position, even with the pointer right over it, so #115's "one of this app's own windows counts as the column" check never actually matched a popover opened from the column, and the grace timer kept firing once the pointer left the column's own window.

Pushed two more changes on top of the `.fixedSize`:
- Fold now checks the popover's own frame directly instead of relying on `windowNumber(at:)` for it.
- While the popover (bookmarks, or the extensions menu, same mechanism) is open, the column is held out regardless of where the pointer wanders, since e.g. clicking a bookmark opens its link in the tab underneath the popover, moving the pointer there without ever closing the popover itself.

Kept the ScrollView `.fixedSize` from before: still harmless, still matches the extensions menu's existing pattern, but the actual fix for #88 turned out to be in `Fold.swift`, not here. Tested by hand with a populated bookmarks list and the hidden-sidebar mode on: hovering into the popover, moving off toward the page, and clicking a bookmark all keep it open until it's actually meant to close.

driceroland:
Thanks a lot @merttopuz, and sorry this one waited: the diagnosis is right, the fold is what takes the popover away, and main still has the bug.

What we'd keep is the heart of it: `ownPopoverOpen` holding the column out while the bookmarks list (or the extensions menu) is open, in both places you put it. Three changes before it goes in:

1. **Keep `onOwnPanel` as it is on main.** The rewrite checks every visible Search window's frame and drops the `top == …` test, so any of our windows under the pointer now keeps the column out even when another app's window is on top of it (or the floating video sits over the column). With the popover hold in place, the rewrite isn't needed for either popover.
2. **Drop the `.fixedSize` in Bookmarks.swift** — as you say yourself, it isn't the fix.
3. **Fold as soon as the list closes:** today the column stays out until the pointer next moves. An `onChange(of: browser.bookmarksOpen)` next to the existing `editingTab` one would settle it right away.

Also, picking a bookmark already closes the popover on main (`pickBookmark`), so that part of the description no longer applies.

If you don't have the time, say so and we'll trim it to that and land it with you as the author.

merttopuz:
Done - all three:

1. `onOwnPanel` back to the topmost-window-number check, as on main.
2. Dropped the `.fixedSize` in Bookmarks.swift.
3. Added `onChange(of: browser.bookmarksOpen)` next to `editingTab` so the column folds the moment the list closes, not on the pointer's next move.

`swift build` passes. Tested by hand with the hidden-sidebar mode on: hovering into the bookmarks popover, moving off toward the page, and letting the list close all behave as expected now.

PR #90, opened 2026-09-23
Fix 60 fps limit on 120 Hz displays

### LINKED ISSUE #75: Bug: Browser appears capped at 60 FPS

Jesper-N:
## Summary

Search limits page animations to around 60 fps, even on a 120 Hz display. This fix turns off WebKit's "Prefer Page Rendering Updates near 60fps" setting so pages can use the display's full refresh rate.

The setting requires a private WebKit API. Search checks that both the API and the setting exist before using them. If either is missing, it keeps WebKit's default behavior.

## Testing

Tested on a 120 Hz Mac running macOS 27.0. A `requestAnimationFrame` test confirmed the increase from about 60 fps to 120 fps.

The project builds with `swift build` using the macOS 26.5 SDK. The build still reports warnings in unchanged code and from the local linker setup.

Closes #75


driceroland:
Thank you @Jesper-N for tracking this down! Three PRs fixed #75 the same way. We're going with #87, which keeps 60 fps by default like Safari (no battery cost unless you turn 120 Hz on), so I'm closing this one. Thanks for the diagnosis.

PR #93, opened 2026-09-23
Site Information: is this connection actually secure?

### LINKED ISSUE #92: Site Information: is this connection actually secure?
A lock icon usually answers this with a click - Search has nothing like it. A tab's right-click menu could offer Site Information…: whether the connection is encrypted and the certificate checks out, who it was issued to, who issued it, and until when.

merttopuz:
Closes #92

A lock icon usually answers this with a click - Search has nothing like it. A tab's right-click menu now has Site Information…, focused on security rather than the fuller privacy/history report some browsers show:

- Whether the connection is encrypted and the certificate checks out - asking the same `serverTrust` WebKit already evaluated to let the page through, not fetching anything anew
- Issued to
- Issued by (the certificate's organisation, or its common name if it gave no organisation)
- Valid until

Site Information panel showing a secure connection with certificate details

## Test plan

- Right-click a tab on an https site, Site Information…: shows secure, issued to, issued by, and the expiry date
- An http site: shows not secure, no certificate fields
- Esc closes the panel, same as every other one
- Disabled on a blank tab, same as Copy Address


driceroland:
Thank you @merttopuz! We've decided to keep Search's menus and panels to the minimum, so I'm closing this one. (If it comes back one day, the trust evaluation shouldn't run in the view body on the main thread.) Thanks!

PR #94, opened 2026-09-23
Allow pages to use full display refresh rate (#75)

### LINKED ISSUE #75: Bug: Browser appears capped at 60 FPS

SamarthaB10:
## Change

- Disable WebKit's near 60 FPS preference before Search creates a page view.
- Apply the setting to normal and private tabs, and to views made from a configuration supplied by WebKit.
- Expose the preference state in `bench probe` and add an Unreleased changelog line.

## Cause

WebKit enables `PreferPageRenderingUpdatesNear60FPSEnabled` on a new `WKPreferences`. Search did not change it, so pages could stay near 60 FPS on a 120 Hz display. WebKit has no public switch for this setting. The change checks for its feature API and uses it only when available.

## Checks

- `./build.sh debug` passes on the latest `main`.
- The runtime probe reported `prefersNear60FPS: [true]` before the fix and `false` after it. On the rebased build, normal, private, and site opened tabs all reported `false`.
- `swift test` builds, then reports `no tests found` because this repository has no test target.
- A live 120 Hz FPS result needs a foreground ProMotion check. This automation desktop changes window focus during frame timing.

Closes #75


driceroland:
Thank you @SamarthaB10! Three PRs fixed #75 the same way; we're going with #87, which keeps 60 fps by default like Safari, so I'm closing this one. I'll carry over your cached flag lookup and the bench probe into #87. Thanks!

PR #99, opened 2026-09-23
The column's scroll bar sits beside the tabs, not over their cross

### LINKED ISSUE #83: Bug: Tab bar scrollbar overlaps the close (×) button
When the tab column has enough tabs to scroll, the scrollbar sits on top of the tab rows and covers the × on the hovered tab. The two are hard to tell apart, and it's easy to grab the scrollbar when you meant to close the tab, or the other way around. 

Image

caisergan:
Closes #83.

With enough tabs to scroll, the column's system scroll bar was drawn over the rows' right edge, which is where the cross on the tab under the pointer sits. The two were hard to tell apart, and it was easy to grab one when aiming for the other.

The column now hides the system bar (as the row across the top already does) and draws its own: a thin line in the margin between the rows and the column's hairline, clear of the cross. Its length and position come from the rows' measured height and scroll offset, and it shows while the pointer is over the rows.

- It takes no clicks: that margin is also the column's edge, which is pulled to resize it, and a bar there would fight it.
- While the tabs fit, the column is the plain stack it was; nothing changes there.
- The tab you go to (⌘1–⌘9, ⇧⌘]) is still scrolled into view.

All of it is in `ThinScroll`, a small view in `Sources/Search/Side.swift`. Measured with `GeometryReader`, so it runs on macOS 14 like the rest of the app. The changelog gets its line under Unreleased › Fixed.

### Testing
- [x] In the app: sidebar layout with more tabs than fit; the bar shows beside the rows on hover, the cross on a row near the bottom closes its tab, and ⌘1–⌘9 scroll the tab into view


driceroland:
Thank you @caisergan! I swapped the custom bar for a lighter trick: the scroll view reaches into the margin and the rows are padded back, so macOS's own scroller lands beside the tabs and nothing is tracked per frame. Merging!

PR #100, opened 2026-09-23
feat: add Universal binary (Intel + Apple Silicon) support

Assim-Genshi:
## Summary

This PR enables building Search as a **Universal 2 macOS application (`arm64` + `x86_64`)**, allowing the app to run natively on both Apple Silicon and Intel Macs. It also fixes an immediate launch crash caused by early `NSApp` access and resolves Swift 6 concurrency warnings and toolchain compatibility issues.

---

### 1. Universal Binary Build (`build.sh` & `Package.swift`)
- Updated `build.sh` to compile separate `arm64-apple-macosx` and `x86_64-apple-macosx` slices and combine them into a single fat binary using `lipo`.
  - *Note:* This avoids the known SwiftPM / Xcode build system bug with duplicate intermediate outputs (`.abi.json`) when attempting to pass multiple `--arch` flags in a single invocation.
- Added explicit `.executable` product declaration in `Package.swift`.
- Added `ARCHS` environment variable override in `build.sh` for fast single-arch local development builds (`ARCHS=$(uname -m) ./build.sh`).

### 2. Fix Early Launch Crash (`Prefs.swift`)
- **Bug:** When launching the app, SwiftUI initializes `@StateObject` properties before AppKit finishes instantiating `NSApplication`. Accessing `NSApp.appearance` in `Preferences.init()` caused an immediate fatal crash (`EXC_BAD_INSTRUCTION`) because `NSApp` is an implicitly unwrapped optional (`NSApplication!`) and was `nil` during early struct allocation.
- **Fix:** Used optional chaining (`NSApp?.appearance = ...`) and dispatched appearance updates to `DispatchQueue.main.async` so that it cleanly applies as soon as the main runloop starts. Added defensive optional chaining in `Icons.swift` and `Design.swift`.

### 3. Toolchain & SDK Compatibility (`Extensions*.swift`)
- Guarded `WKWebExtension` APIs (which require macOS 15.4 / Xcode 16.3 SDK) behind `#if compiler(>=6.1)` with lightweight fallback stubs for older toolchains.
- The project now builds cleanly across both macOS 14 / Xcode 16.2 and newer toolchains without breaking forward compatibility.

### 4. Swift 6 Concurrency & Warning Cleanups
- **`History.swift` & `Icons.swift`:** Marked static file/folder properties `nonisolated` and captured them locally before dispatching to background utility queues.
- **`Icons.swift`:** Wrapped background `NSImage` decoding in an `@unchecked Sendable` struct (`SendableImage`) to eliminate actor boundary transfer warnings.
- **`Links.swift`:** Properly annotated the `onceShown` completion handler with `@Sendable @MainActor` and used `MainActor.assumeIsolated`.
- **`Bench.swift` & `Browser.swift`:** Silenced unused `fcntl` return value warnings and added `@preconcurrency import WebKit`.

---

## Verification

1. **Universal Binary Verification:**
   ```bash
   $ lipo -info build/Search.app/Contents/MacOS/Search
   Architectures in the fat file: build/Search.app/Contents/MacOS/Search are: x86_64 arm64


driceroland:
Thank you @Assim-Genshi! We'd like Intel support later, but not by doubling the app's size for everyone for now: Search's small download is part of what it is, and the release chain (Homebrew cask, appcast) would need a separate Intel build. When we get there, a PR limited to build.sh would be the way. Native architecture for debug, universal only in release, without the toolchain stubs. So I'm closing this one for now. Thanks!

Issue #104, opened 2026-09-24
New tab at the foot of the sidebar, as an option


PeterTheMango:
Part of #96. Lives on the Customization page (#101). Related: #102, #103.

In the sidebar, the **New tab** row sits right below the last tab, so its position changes as tabs open and close. Once the column scrolls, it scrolls away with the tabs. I'm used to having New tab in a fixed place at the foot of the sidebar, so I can reach it without looking for it.

**Proposal:** a switch that moves New tab to the foot of the sidebar, as a **+** in its bottom right corner, drawn like the other buttons there (the bookmarks, and the space and extensions when those are shown). The row under the tabs goes away while it's on.

- **Off by default:** New tab stays under the last tab.
- Shown only when Tabs in a sidebar is on. The row across the top isn't affected.
- Open to other places for it, e.g. the top of the column above the tabs, where Arc keeps it.

### Preview

<img width="530" height="152" alt="The + at the foot of the sidebar" src="https://github.com/user-attachments/assets/2142ccbd-7a64-423a-9c43-5092d0c10af8" />

Issue #103, opened 2026-09-24
The zoom level at the top of the page, as an option


PeterTheMango:
Part of #96. Lives on the Customization page (#101). Related: #102, #104.

When I zoom a page (⌘+ / ⌘− / pinch), the percentage appears in the capsule that rises from the bottom of the window, the same one that says "Address copied." I don't look at the bottom of the window while zooming, so I usually miss it. Chrome and Firefox show the zoom level at the top right, next to the address, so that's where I look first.

**Proposal:** a **Show zoom at** dropdown: **Bottom middle** (today, the default), **Top middle**, **Top left** or **Top right**. The zoom level uses the same capsule and the same fade in whichever spot is picked.

- The spots are measured from the page, not the window, so they stay clear of the sidebar and, with tabs across the top, sit just below the tab row. At the top right it moves down out of the find bar's way while that is open.
- Only the zoom level moves. "Address copied" and the other messages stay at the bottom.

### Preview

<img width="800" height="556" alt="The zoom level at the top of the page" src="https://github.com/user-attachments/assets/5618cbb3-2024-4be9-b8da-ca4b2e38f841" />

Issue #102, opened 2026-09-24
Tabs a size bigger, as an option


PeterTheMango:
Part of #96. Lives on the Customization page (#101). Related: #103, #104.

Tab titles are 12.5 pt and a sidebar row is 28 pt tall, in both the sidebar and the row across the top. That's too small for some people to read quickly. I find it easier to spot the tab I want when the tabs are a bit bigger.

**Proposal:** a **Tab size** setting with two choices, **Regular** (today, the default) and **Large**. Large makes the title, the site's icon and the close button bigger and the rows taller to fit, e.g. 14 pt titles and 34 pt rows. Pinned squares grow with them.

- Only two sizes, no slider. Nobody sees a change unless they pick Large.
- **How others do it:** Firefox has a density setting (Compact / Normal / Touch) that changes the tab height. On the Mac, Finder and Mail sidebars follow System Settings › Appearance › Sidebar icon size.

### Preview

<img width="241" height="800" alt="Tab size, Regular and Large" src="https://github.com/user-attachments/assets/cb8c3f57-9f4d-4dd2-8767-09d6d540f207" />

Issue #96, opened 2026-09-23
Customization Settings Tab


PeterTheMango:
> **Split into separate issues**, each with its own pull request, so they can be discussed and merged one at a time:
>
> - #101 Settings › Customization page (PR #105)
> - #102 Tab size (PR #106)
> - #103 Where the zoom level shows (PR #107)
> - #104 New tab at the foot of the sidebar (PR #108)

---

A **Customization** page in Settings would give people one place to change how their Search looks and feels.

Today these settings are spread out. The Light / Dark / System switch sits on the General page next to spelling, middle-button scrolling and the bench socket. How tabs look is on the Tabs page, mixed in with sleeping tabs and Spaces. Several open proposals would add more options of this kind, and each would need a place in Settings:

- An address bar above the page (#15, #56)
- A tab row in the page's colour (#25)
- Bookmarks in the column, above the tabs (#58, #69)

A Customization page, between General and Tabs, could be the home for all of these. It would start with the Light / Dark / System switch, moved from General as it is, and the three options below. Each new option defaults to how Search works today, so nobody sees a change unless they turn it on.

### 1. Tab size
Tab titles are 12.5 pt and a sidebar row is 28 pt tall, in both the sidebar and the row across the top. That's too small for some people to read quickly. I find it easier to spot the tab I want when the tabs are a bit bigger.

- **Tab size:** **Regular** (today, the default) or **Large**. Large makes the tab's title and icon bigger and makes rows taller to fit, e.g. 14 pt titles and 34 pt rows. Pinned squares grow with them.
- Only two sizes, no slider.
- **How others do it:** Firefox has a density setting (Compact / Normal / Touch) that changes the tab height. On the Mac, Finder and Mail sidebars follow System Settings › Appearance › Sidebar icon size.

### 2. Where the zoom level shows
When I zoom a page (⌘+ / ⌘− / pinch), the percentage appears in the capsule that rises from the bottom of the window, the same one that says "Address copied." I don't look at the bottom of the window while zooming, so I usually miss it. Chrome and Firefox show the zoom level at the top right, next to the address, so that's where I look first.

- **Show zoom at:** **Bottom middle** (today, the default), **Top middle**, **Top left** or **Top right**, chosen from a dropdown or a small picker showing the four spots on a window outline. It uses the same capsule and the same fade in whichever spot is picked. The spots are measured from the page, not the window, so they stay clear of the sidebar and, with tabs across the top, sit just below the tab row.
[...]

PR #105, opened 2026-09-24
Settings › Customization: a page for how Search looks, with the appearance moved there

### LINKED ISSUE #101: Settings › Customization, a page for how Search looks
Part of #96, split out so each piece can be discussed and merged on its own.

Settings has no single place for how Search looks. The Light / Dark / System switch sits on the General page next to spelling, middle-button scrolling and the bench socket, and several open proposals would add more options of this kind (an address bar above the page #15, a tab row in the page's colour #25, bookmarks in the column #58).

**Proposal:** a **Customization** page in the Settings sidebar, between General and Tabs. It starts with the Light / Dark / System switch, moved from General as it is, and becomes the home for the options that change how Search looks:

- #102 Tab size
- #103 Where the zoom level shows
- #104 New tab at the foot of the sidebar

This page doesn't change what Search does or any default; every setting keeps its value.

### Preview

With the tab size and zoom options from the linked issues in place:

Settings › Customization

PeterTheMango:
Closes #101 (part of #96)

Settings gets a **Customization** page, between General and Tabs, for how Search looks as opposed to what it does. It starts with the Light / Dark / System switch, moved over from General as it is: same setting, same saved value, same behaviour. The options in #102, #103 and #104 go on this page, each in its own PR on top of this one.

Settings › Customization

*The preview is taken with the next two PRs in place; on its own, this PR's page has only the Appearance line.*

### How
- `SettingsPanel.Page` gets `.customization` (icon `paintbrush`), and the Appearance line moves from `general` to a new `customization` card.
- The three comments that pointed to "Settings › Appearance" (App.swift, Design.swift, Tab.swift) now say Settings › Customization › Appearance.
- CHANGELOG line added.

### Testing
- Settings › Customization shows Appearance; General no longer does. Switching Light / Dark / System works as before, and a saved choice carries over.


driceroland:
Thank you @PeterTheMango! We've decided to keep Settings as minimal as possible, so I'm closing the Customization page along with the options built on it. Thanks for the clean work.

PR #106, opened 2026-09-24
Tabs can be a size bigger: Settings › Customization › Tab size

PeterTheMango:
Closes #102 (part of #96)

> Builds on #105 (the Customization page), so this PR's first commit is that one. Only the last commit, "Tabs can be a size bigger", is new here.

Settings › Customization gets **Tab size**: **Regular**, the size tabs have always been and the default, or **Large**. Large grows the title, the site's icon, the close button (and the loading ring and speaker that share its place), the row height and the pinned squares, in the sidebar and in the row across the top.

Tab size, Regular and Large

### How
- `TabSize` (Prefs.swift) holds the numbers for each size: title 12.5 → 14 pt, mark 15 → 17, sidebar row 28 → 34, pinned square 34 → 40, close button 15 → 18, and in the row across the top a tab 186 → 208 wide and a pin 30 → 34, so a larger title isn't just less of it. Saved as `tabs.size`, default `regular`.
- `SideBar`'s fixed `row` and `square` become the chosen size's; the pinned squares already scale everything inside off their height, so they follow on their own. `SideRow`, `TabPill` and the New tab row read their sizes from it too.
- `TabBar` already redraws on any preference change (Browser forwards `prefs.objectWillChange`), so the row's widths follow at once.
- CHANGELOG line added.

### Testing
- Regular: tabs look exactly as before, in both layouts.
- Large, in the sidebar and across the top: titles, icons, close button and rows grow; pinned squares grow with them; switching back and forth animates in place.
- `swift build`: no new warnings.

PR #107, opened 2026-09-24
The zoom level can show along the top of the page: Settings › Customization › Show zoom at

### LINKED ISSUE #103: The zoom level at the top of the page, as an option

PeterTheMango:
Closes #103 (part of #96)

> Builds on #106 (tab size), which builds on #105 (the Customization page). Only the last commit, "The zoom level can show along the top of the page", is new here.

Settings › Customization gets **Show zoom at**, a dropdown: **Bottom middle** (where it is today, and the default), **Top middle**, **Top left** or **Top right**. The zoom level keeps its capsule and its fade; only where it shows changes.

The zoom level at the top of the page

### How
- `ZoomSpot` (Prefs.swift), saved as `zoomspot`, default `bottom`. Not under `zoom.`, where each site's own zoom level is kept by its host.
- At the bottom nothing changes: the zoom level still goes through `announce`. Anywhere else it goes to a `zoomNote` of its own, on the same 1.7 s clock, so "Address copied" and the rest still show at the bottom instead of being pushed aside.
- The capsule is drawn as an overlay of the page, not the window, so it stays clear of the sidebar and below the tab row. At the top right it moves down out of the find bar's way while that is open.
- The capsule's look is shared with the bottom line through one `note(_:)`.
- CHANGELOG line added.

### Testing
- Bottom middle: ⌘+ / ⌘− / pinch read out at the bottom exactly as before.
- Each of the three top spots, with the sidebar and with tabs across the top: the level shows in that spot and fades; with ⌘F open, the top right sits below the find bar; ⌘⇧C still says "Address copied" at the bottom.


driceroland:
Thank you @PeterTheMango! We've decided to keep Search's settings to the minimum, so I'm closing the zoom position option (and the Customization page with it). Thanks for the clean work.

PR #108, opened 2026-09-24
New tab can sit at the foot of the sidebar, off until turned on

### LINKED ISSUE #104: New tab at the foot of the sidebar, as an option

PeterTheMango:
Closes #104 (part of #96)

> Builds on #107, #106 and #105. Only the last commit, "New tab can sit at the foot of the sidebar", is new here.

Settings › Customization gets **New tab at the foot of the sidebar**, off by default. On, the New tab row under the last tab goes, and a **+** takes its place in the bottom right corner of the sidebar's foot, where it doesn't move as tabs come and go or scroll with them. The switch shows only while tabs are in the sidebar; the row across the top has its own + and isn't affected.

The + at the foot of the sidebar

### How
- `Preferences.newTabInFoot`, saved as `sidebar.newtab.foot`, default off.
- `SideBar.foot` gets a `Door(icon: "plus")` after its spacer; `rows` and the other spaces' previews leave out the New tab row while it's on, and `rowsEnd` (where the window's drag area starts) stops counting it.
- CHANGELOG line added.

### Testing
- Off: the sidebar is exactly as before.
- On: the + sits in the bottom right corner, with Spaces on and off, at both tab sizes; it opens a new tab; the row under the tabs is gone, and comes back when switched off.


driceroland:
Thank you @PeterTheMango! We're keeping Search's options to the minimum, so I'm closing this one. Thanks!

PR #109, opened 2026-09-24
A tab's title in the column runs to the end of its row

### LINKED ISSUE #84: Tab titles cut off early to leave room for a hidden × button
Tab titles in the column are truncated well before the edge of the row, leaving a wide empty gap between the end of the text and the ×. The gap is there even on tabs that aren't hovered, where the × isn't shown at all, so every title loses room it could use.

Image

caisergan:
Closes #84.

Every row in the column kept a place at its end for the cross, empty unless the pointer was on the tab. With the spacing around it, a title stopped about 40pt short of the row's edge, against 10pt at the start, so every title lost room it could use.

Now the title runs to the end of its row, 10pt from the edge as from the start. Under the pointer the cross draws over the end of the row, as before in the same place and with the same click area, and the title fades out beneath it instead of being cut shorter. That way the end of the title holds still as the pointer passes down the column, rather than jumping on each row.

- The loading ring and the speaker stay for as long as the page loads or plays, so a tab showing one still keeps its place at the end; under the pointer the cross takes it.
- Editing a tab's address in place is as it was.
- The row across the top (`TabPill` in `TabBar.swift`) keeps its place for the cross the same way; left as it is here, since the issue is about the column.

All of it is in `SideRow` in `Sources/Search/Side.swift`. The changelog gets its line under Unreleased › Fixed.

### Testing
- [x] In the app: sidebar layout with long titles; at rest they run to the end of the row, under the pointer the cross shows and closes the tab, and the title fades beneath it


driceroland:
Merged, thank you! Checked together with today's other PRs in a combined build, in a hidden test world: it builds and the column draws; the hover look is for Drice's eyes.

PR #110, opened 2026-09-24
Report the installed Safari version in the user agent

peter-bf:
The user agent hardcoded Version/26.5, but web views run on the system's WebKit, which can be older (for example, Safari 18.6 on macOS 15 instead of the hardcoded 26.5). 

Sites that choose code paths by browser version could serve features the engine does not support and bot protection that compares the claimed version against the engine's actual features may treat the mismatch as suspicious traffic.

Read CFBundleShortVersionString from the installed Safari at launch, checking the cryptex location first and then /Applications. If neither is readable, fall back to the Safari version that shipped with the running macOS, the oldest the system WebKit can be.


driceroland:
Merged, thank you! Checked together with today's other PRs in a combined build, in a hidden test world: navigator.userAgent says Version/26.6.2, the Safari installed on that Mac.

PR #113, opened 2026-09-24
Brazilian Portuguese: the app's own UI in pt-BR

Luizhcrs:
## Summary

- Adds a `pt-BR` localization of Search's own interface — every menu, Settings, History, Bookmarks, Passwords, Extensions, the welcome screens, and every dialog — alongside the existing English.
- A one-line `L(_:)` helper in `Sources/Search/Localization.swift` wraps `NSLocalizedString` against `Bundle.main`; call sites elsewhere only needed wrapping, no logic changes.
- `Localization/en.lproj/Localizable.strings` and `Localization/pt-BR.lproj/Localizable.strings` hold the two tables (491 keys each). `build.sh` now copies both `.lproj` folders into `Contents/Resources`, and `Info.plist` declares `CFBundleLocalizations` (`en`, `pt-BR`) and `CFBundleDevelopmentRegion`. Search then follows whatever language System Settings › General › Language & Region is set to, the same as any other Mac app — no in-app language switch to build or maintain.
- Left untouched on purpose: page content and any JS/CSS injected into sites, other browsers' and services' own names (Chrome, Google, etc.), and Search's own name.
- One correctness fix that fell out of this: `CommandMenu(L("Bookmarks"))`'s title changes per language, but three call sites (`Bench.swift`, `Bookmarks.swift` ×2) found that menu by comparing `$0.title == "Bookmarks"` — which would silently stop matching in `pt-BR`. They now compare against `L("Bookmarks")` too.

## Test plan

- [x] `swift build` — clean, no warnings introduced.
- [x] `./build.sh` — produces `build/Search.app` with both `.lproj` folders in `Contents/Resources` and both languages listed in `Info.plist`.
- [x] Launched the built app with `-AppleLanguages "(pt-BR)"` and read the live menu bar via the Accessibility API: app menus and system-provided ones (File/Edit/View/Window/Help) all switch — `Arquivo, Editar, Visualizar, Abas, Favoritos, Histórico, Janela, Ajuda` — and items inside them (e.g. the Abas menu: `Voltar, Avançar, Próxima Aba, Aba Anterior, Pesquisar Abas…, Fixar Aba, Renomear Aba, Duplicar Aba…`) read correctly.
- [ ] A native Brazilian Portuguese speaker's read-through of the translations would be welcome — I did my best on tone and macOS/browser convention, but I'm not a native speaker.


driceroland:
Obrigado @Luizhcrs! This is careful work and safe (Bundle.main, no SwiftPM resources). We've decided not to localize Search for now: every new string would then need a translation, and the app is still changing fast. I'm closing it, but thank you very much.

PR #115, opened 2026-09-24
A hidden sidebar opens at the edge and closes when you leave it, every time

### LINKED ISSUE #114: Bug: Hidden sidebar sometimes won't open at the left edge, or won't close after leaving it
With "Hide the sidebar until the pointer reaches the edge" on, the column doesn't always follow the pointer:

  - **Won't open:** sometimes moving the pointer to the left edge doesn't bring the sidebar out. It comes back after moving away from the edge and trying again.
  - **Won't close:** after bringing the sidebar out and letting it go quickly 5-10 times, it stayed out after the pointer had left it, until the pointer
  went into the column and out again.

  ### Steps
  1. Settings › Tabs: turn on Tabs in a sidebar and Hide the sidebar until the pointer reaches the edge.
  2. Window maximized, flush against the left edge of the screen.
  3. Move the pointer to the left edge and away again, quickly, several times.

  ### Expected
  The sidebar comes out every time the pointer rests on the left edge, and goes every time the pointer leaves it.

  ### What happens
  After a few quick in-and-outs, either the edge stops bringing the sidebar out, or the sidebar stays out with the pointer elsewhere.

caisergan:
Closes #114.

With "Hide the sidebar until the pointer reaches the edge" on, the column didn't always follow the pointer. After opening and closing it quickly a few times, either the left edge stopped bringing it out, or it stayed out after the pointer had left it.

The cause was the hover events `Fold` relied on: SwiftUI's `onHover` over a 6pt band at the edge to open, and over the column itself to close. A view that appears under a pointer that isn't moving never gets "entered", so it never gets "exited" either. The column coming back under the pointer never closed, and the band could think the pointer was still on it and ignore it coming back.

`Fold` now reads the pointer's actual position on every move, through local and global `NSEvent` monitors, and compares it with the window:

- At the left edge (the top edge for the strip), it starts the dwell and opens.
- Off the column once it's out, it starts the grace and closes. The grace now counts from when the pointer left, not from its latest move.
- Another app's window over the edge doesn't open it. One of Search's own windows, such as a popover from the column, counts as part of the column.
- It works with another app in front too.

The strip folded across the top used the same code and gets the same fix.

Side effects:
- The edge bands are gone, so the page's left 6pt takes clicks again.
- The main window now accepts mouse-moved events (`acceptsMouseMovedEvents`), so the monitor sees moves anywhere over it.

All of it is in `Sources/Search/Fold.swift`. The changelog gets its line under Unreleased › Fixed.

### Testing
- [x] In the app, window maximized: opened and closed the sidebar quickly many times; it opened at the edge and closed on leaving every time


driceroland:
Thank you @caisergan, this fixes a real annoyance. One change: the mouse monitors are now installed only while the sidebar or top bar is actually folded, `acceptsMouseMovedEvents` is only on meanwhile, and `windowNumber(at:)` is only asked near the edge or while the column is out, so nothing runs on every mouse move otherwise. The #114 fix and the removal of the click-taking strip are kept. Merging!

PR #117, opened 2026-09-24
An extension worker's WebSockets are opened by the browser

### LINKED ISSUE #116: 1Password never finishes signing in: its worker freezes on new WebSocket()
**What I did**

Installed 1Password from the Chrome Web Store (stable 8.12.37, later Nightly 8.12.42) and signed in, both through the 1Password Mac app and through the "Sign in" page on my.1password.com. The Mac app refuses Search until it's added under 1Password › Settings › Browser › Add Browser, so I did that first.

**What I expected**

The extension picks up the account and unlocks.

**What happened**

The popup never gets past its loading spinner, and the my.1password.com sign-in tab spins forever. The extension's own log (its `b5x-diagnostics` IndexedDB) stops right after the Mac app hands it a session:

```
[AppIntegration] [DesktopApp] Initiation complete - B5X is connected to desktop app
[DelegatedSessions] Signing in with delegated session
[AppIntegration] Received message <NmRequestDelegatedSession> from native core. Duration: 274ms
```

Nothing after that, and no account gets saved. Search notices and restarts the extension ("its worker stopped answering"), but the new one never starts either.

**Why**

`sample` on the WebContent process that hosts the worker gives the same stack in every sample:

```
WebCore::WorkerMainRunLoop::postTaskForMode
  WebCore::EventLoop::run
    WebCore::MicrotaskQueue::performMicrotaskCheckpoint
      WebCore::constructJSWebSocket1
        WebCore::WebSocket::create → WebSocket::connect
          WebCore::WorkerThreadableWebSocketChannel::WorkerThreadableWebSocketChannel
            WebCore::waitWithSTWParticipation
              WTF::BinarySemaphore::waitUntil
```

WebKit runs an extension's service worker on the main thread of its process (`WorkerMainRunLoop`). A WebSocket made from a worker goes through `WorkerThreadableWebSocketChannel`, which sets the connection up on the main thread and waits for it. The worker is the main thread, so it waits on itself. It never runs again, and a restarted extension gets handed the same stuck worker.

For 1Password the socket is its notifier (`this.websocket = new WebSocket(e)` in `@1password/web-api/notifier.js`), opened as soon as the session is set up.

It reproduces without 1Password. A worker with only this in it never replies, and nothing else gets through to it after that:

```js
chrome.runtime.onMessage.addListener((m, sender, reply) => {
  const ws = new WebSocket("wss://echo.websocket.org");
  ws.onopen = () => reply("open");
  return true;
});
```

This might also be the Bitwarden item on the [roadmap](https://github.com/driceroland/Search/blob/bbc0c6e1f580cae01a2db853a790125fc8bf0fc0/ROADMAP.md?plain=1#L14) ("goes blank after signing in"). Bitwarden's worker connects to its notifications hub over SignalR once you're signed in, with `transport: WebSockets, skipNegotiation: true` in its `background.js` (2026.8.0). I don't have a Bitwarden account, so I haven't checked.

macOS 27.2, Search 1.0.1.

anyrobert:
Fixes #116.

It might also fix the Bitwarden item on the [roadmap](https://github.com/driceroland/Search/blob/bbc0c6e1f580cae01a2db853a790125fc8bf0fc0/ROADMAP.md?plain=1#L14) ("goes blank after signing in"), since Bitwarden's worker opens a SignalR WebSocket once you're signed in. I couldn't test that without an account, so I left the roadmap line alone.

**What it does**

The details of the freeze are in #116: WebKit runs an extension's service worker on the main thread, and a WebSocket made from a worker waits on the main thread, so it waits on itself.

The shim now replaces `WebSocket` in extension workers only. The replacement opens a `connectNative("search.websocket")` port, and `ExtensionSocket.swift` makes the actual connection with `URLSessionWebSocketTask`, passing text and binary frames both ways. Extension pages keep WebKit's own WebSocket, which works there.

The request is the one the worker would have made: same URL and subprotocols, `Origin: chrome-extension://<id>`, the worker's user agent, and whatever cookies Search's store has for that host. The same connection to the same server, just started from the app instead of WebKit, so nothing new leaves the app.

One thing that surprised me: WebKit drops messages posted on a native port before the delegate has set its `messageHandler`, and the worker posts right after `connectNative()`. Without a handshake, most connections never opened. So Search replies `ready` and `opening`, and the worker re-sends its `open` every 250 ms until it hears back, giving up after 10 s as if the connection had been refused.

**Testing**

- A small test extension whose worker opens `wss://echo.websocket.org`, sends text and binary, then closes. On `main` the worker freezes in `new WebSocket` with the same stack as in #116. With this change: open, both echoes, close 1000 clean, and the worker still answers messages afterwards. Nine runs in a row, all the same.
- An unknown host and a refused port: `error`, then `close` 1006.
- Four sockets at once: three closed cleanly, and the fourth got a 1006 because echo.websocket.org turned its handshake down (`-1011`), which is what a browser reports too.
- 1Password Nightly 8.12.42 with a real account: signs in, syncs, and keeps working, where it used to stop right after getting its session.


driceroland:
Thank you @anyrobert, and for #116: your diagnosis of the freeze was spot on. #126 fixed the same WebSocket freeze in the same way (a native port, with URLSession on the other end) along with 1Password's other two problems, and it's now merged on main, so I'm closing this one as covered. I also fixed one thing both PRs share: what a worker posts on a port it has only just opened is often dropped, so the opening is now said again on the same port until Search answers. Your note that Bitwarden opens a SignalR socket once signed in is probably the real cause of its blank popup; we'll check that with a real account on the next build. Thanks again.

PR #119, opened 2026-09-24
Choose how quickly the hidden sidebar opens

### LINKED ISSUE #118: Option to select sidebar opening delay
With **Hide the sidebar until the pointer reaches the edge** on, the column waits 0.15 s with the pointer on the left edge before it slides out, then takes about a third of a second to slide in. The wait is there so a pointer passing the edge on its way to the Dock doesn't bring it out by accident.

**Proposal:** a **Show the sidebar** setting in Settings › Tabs, right under Hide the sidebar until the pointer reaches the edge and shown only while that is on, with three choices:
  - **Cheetah**: no wait, and a quicker slide out.
  - **Human**: the 0.15 s wait it has today. The default.
  - **Turtle**: 0.4 s, for a pointer that crosses the edge often on its way somewhere else.

  Each has an SF Symbol beside its name (hare, figure.walk, tortoise); 
  - The default setting is set to 0.15 s. Delay doesn't change unless user changes it from the settings.
  - Only the sidebar. The strip across the top keeps its wait, since its edge is the way to the menu bar.

Image

caisergan:
Closes #118.

With **Hide the sidebar until the pointer reaches the edge** on, the column waited a fixed 0.15 s on the left edge before sliding out, then took about a third of a second to slide in. Settings › Tabs gets a **Show the sidebar** row right under that switch, shown only while it is on, with three choices:

- **Cheetah**: no wait, and a quicker slide out (`Motion.quick` instead of `Motion.glide`).
- **Human**: the 0.15 s wait it had before. The default, so nobody sees a change unless they pick another.
- **Turtle**: 0.4 s, for a pointer that crosses the edge often on its way to the Dock.

How it's put together:
- `Reveal` in `Prefs.swift`, next to `Glyph`: each choice's name, symbol and wait. Stored as `sidebar.reveal`.
- `Fold.swift` waits for the chosen time before bringing the column out; the strip across the top keeps its own 0.15 s, since its edge is the way to the menu bar.
- `Segmented` takes an optional SF Symbol per choice (`hare`, `figure.walk`, `tortoise`; there is no cheetah symbol). Look, Tabs show, Welcome and the new space card, which already use it, are unchanged.
- No new dependencies. The changelog gets its line under Unreleased › Added.

This and #115 both change `arrive()` in `Fold.swift`. Whichever is merged second needs a small rebase; I've already combined the two to test them together and can push that once the first is in.

### Testing
- [x] In the app: each choice from the left edge; Cheetah comes out at once, Human as before, Turtle after a pause; the row only shows while Hide the sidebar is on


driceroland:
Thank you @caisergan! I'm closing this one: we'd rather not add a setting for this delay, and keep Settings minimal. If 0.15 s feels wrong, let's change the default instead. #115 is the fix for the edge that stops answering, and I'll take care of that one.

PR #120, opened 2026-09-24
Right-click "Search with …" searches in Search instead of opening default browser

danielferrazs:
When text on a page was selected and "Search with …" was chosen from the context menu, WebKit invoked macOS's system search service (`NSPerformService`), which handed the search query over to whichever browser was configured as default on the Mac (typically Safari) rather than searching in Search itself.

### What it does

- `PageView.willOpenMenu`: Intercepts the search menu item (`WKMenuItemIdentifierSearchWeb`), re-targets it to `#selector(searchWithEngine(_:))`, and dynamically updates the menu item's title to match Search's configured search engine (e.g. "Search with Google", "Search with DuckDuckGo", or localized equivalents like "Pesquisar com …").
- Captures the selected text via JavaScript (including inside inputs/textareas and iframe frames) and opens the query in a new tab inside Search via `browser.open(url, foreground: true)`, preserving the current window, space, and privacy mode.
- Updates `CHANGELOG.md` under `Unreleased` › `Fixed`.

### Testing

- [✓] `./build.sh`: builds clean, 0 new warnings.
- [✓] Tested in the app:
  - Selecting plain text on web pages and choosing "Search with Google" opens a new tab in Search with the search results, keeping focus in Search without launching Safari.
  - Selecting text inside text inputs and textareas searches the selected text in a new tab.
  - Verified with different search engines configured in Settings (Google, DuckDuckGo, etc.) to ensure the query URL and menu label match the user's setting.


driceroland:
Thank you @danielferrazs! #73 fixes the same thing, and keeps private tabs private (this one opens the search in an ordinary tab from a private one), so I'm closing this one in its favour. I'll try to carry over your idea of reading the selection from same-origin iframes and text fields.

PR #122, opened 2026-09-24
A new tab from a private tab is private too

### LINKED ISSUE #121: Incognito Consistency
When in a private tab, if you open a new tab from a private tab, it opens it in a public tab rather than keeping it private as well. Will continue testing and proposing a fix.

PeterTheMango:
Closes #121

A tab opened from a private tab is now private too. Before, ⌘T, ⌘-click on a link, ⌘D, Open Image in New Tab and a bookmark opened with ⌘ all made an ordinary tab, even from a private one. That tab went into the history and tomorrow's session. Only links that ask for a new window (`target=_blank`, `window.open`) were already kept private.

### How
- `newTab()`: on a private tab, it hands off to `newShyTab()`. So ⌘T, the menu's New Tab and the + in the sidebar and in the row across the top all open a private tab there. The rule of never having two empty tabs still applies, through `newShyTab()`'s own reuse of an empty private tab.
- `open(_:foreground:atEnd:from:)`: a new `from` argument, the tab it was opened out of. A private one's opens private, in **the same** non-persistent store, so a link followed out of a site you signed in to privately is still signed in. That matches what `createWebViewWith` already does with the configuration WebKit hands over. Extension pages keep their own configuration.
- `from` is passed by ⌘/middle-click on a link, ⌘D (and Duplicate in the tab's menu), Open Image in New Tab, and `visit` (a bookmark opened with ⌘ held).
- Left public: tabs an extension opens, links from other apps, History menu items and links dropped on the tab row. None of these come out of a page.
- `Web.configuration` takes an optional `store`, and `Tab.store` exposes its `websiteDataStore`.
- CHANGELOG line added.

### Behaviour worth a look
While a private tab is in front, ⌘T and + can't make an ordinary tab. You switch to an ordinary tab first, the way a Chrome incognito window works. If ⌘T should stay ordinary and only links should stay private, that's the first block of `newTab()` taken out.

### Testing
- Driven end to end in the app: from a private tab (⌘⇧N), the new tabs opened by ⌘T, ⌘-click on a link and ⌘D are private, and none of their pages appear in History.


driceroland:
Merged, thank you! Checked together with today's other PRs in a combined build, in a hidden test world: ⌘T from a private tab, then a page: kept out of the session and the history; the same from an ordinary tab is recorded.

PR #124, opened 2026-09-24
Addresses that redirect, like youtube.com, open the page instead of downloading a file

danielferrazs:
When navigating to an address like `youtube.com`, the server returns an HTTP 301 redirect to `https://www.youtube.com/`, but includes a header of `Content-Type: application/binary`.

Because WebKit cannot natively render that MIME type, `response.canShowMIMEType` evaluated to `false`. Without inspecting the HTTP status code, Search interpreted the response as an unrenderable file download and initiated a `WKDownload`, saving a `.html` file into the user's Downloads folder instead of following the redirect to the site.

### What it does

- In `Browser.webView(_:decidePolicyFor response:)`, any HTTP redirect response (status code in `300...399`) is now unconditionally `.allow`ed so WebKit follows the `Location` header to the destination resource.
- Legitimate file downloads (DMGs, ZIPs, etc.) continue to download as expected because they are served with a `200 OK` status and a non-displayable MIME type.
- Adds an entry to `CHANGELOG.md` under `## Unreleased` › `### Fixed`.

### Testing

- [✓] `./build.sh`: clean build, 0 new warnings.
- [✓] In the app:
  - Typing `youtube.com` into the omnibox now navigates straight to YouTube without triggering a file download.
  - Tested file downloads with redirects (e.g. GitHub release assets) to verify that redirect chains leading to real file downloads still trigger the download when reaching the 200 OK file response.


PeterTheMango:
Thought I was the only one that was experiencing this. Nice

driceroland:
Merged, thank you! Checked together with today's other PRs in a combined build, in a hidden test world: youtube.com now opens www.youtube.com (title YouTube), and nothing lands in Downloads.

PR #125, opened 2026-09-24
⌘W no longer bounces between pins: a pin already put down stays down

armin-ahmadii:
With two pinned tabs and no other tab open, ⌘W bounces between them: ⌘W on A lands on B, and ⌘W on B wakes A again, back and forth, so you can never get out of the pins.

The pin branch of `close()` already avoids this when an ordinary tab exists ("⌘W bounced between the two instead of getting you out of them"). With only pins, the last fallback still picked a pin that had just been put down and woke it. This removes that fallback: once no other tab is awake, closing a pin leaves an empty tab through the `newTab()` that was already there, as closing the last tab of all does.

**Before:** two pins, ⌘W, ⌘W, ⌘W… → A, B, A, B…
**After:** two pins, ⌘W, ⌘W → B, then an empty tab; ⌘W again closes the window as before.

Tested with `./build.sh`:
- two pins, no other tabs: ⌘W twice → empty tab, then the window closes
- pins + an ordinary tab: ⌘W on a pin → back to that tab (unchanged)

**Testing**

[✓] `swift build`: clean build, 0 new warnings.
[✓] In the app (`./build.sh`):
Two pinned tabs, no other tabs: ⌘W on one pin lands on the other, and ⌘W again now leaves an empty tab instead of waking the first pin, so it no longer bounces between them. ⌘W on the empty tab closes the window as before.
Pins plus an ordinary tab: ⌘W on a pin still goes back to the ordinary tab, unchanged.


driceroland:
Merged, thank you @armin-ahmadii! Tested in a hidden test build with two pins awake: ⌘W on the first lands on the second, ⌘W on the second now leaves an empty tab and the first stays put down, no more bouncing. It's in the changelog for the next update.

PR #127, opened 2026-09-24
Prevent webpage hover behind full-page panels

luca-dupont2:
## Problem

A web page can keep its CSS hover state when Search opens Settings over it. I reproduced this on Search 1.0.1 and macOS Tahoe 26.6.2 with a local page that marks hovered elements. The same native overlay path is used by History, Downloads, Bookmarks, and Passwords; the Hidden list, Welcome panel, and address field can cover the page too.

## Change

Give WebKit a transparent pointer target in the active page while a full-page overlay is visible. Remove it when the overlay closes. Reapply it after navigation and when the active tab changes. This keeps page hover from showing through Search's panels without adding a dependency or a preference.

## Verification

- Before the change, the local page kept a hovered element after Settings opened.
- With the final build, moving the pointer inside Settings or History left no site element hovered. Closing the overlay removed the guard and restored normal hover.
- Checked the address field, tab switches while Settings was open, and navigation while Settings was open.
- Swift build succeeded with no new warnings. Existing deprecation and unused-result warnings remain on unchanged lines.


driceroland:
Thank you @luca-dupont2! I'm closing this one in favour of #149, which fixes the same cause natively. Writing an element into the page's own DOM is visible to the site and can be lost when it rewrites the document. Thanks for spotting it.

PR #128, opened 2026-09-24
Extensions: hold the browsing APIs to what the manifest asked for

Hyp4tia:
## What

The layer that fills in the Chrome APIs WebKit lacks answers from the browser itself: bookmarks, history, downloads, the open tabs, closed tabs, top sites, browsing data, the reading list. Until now the permission checks for those answers lived only in the injected script (`search-shim.js`), which runs beside the extension's own code, while `Extensions.swift:250` hands every extension `nativeMessaging` at every load and `runtime.sendNativeMessage("search", …)` reaches `ExtensionShims.answer` directly (`Extensions.swift:868-871`). An extension that skips the script can ask the browser itself.

Demonstrated on `bbc0c6e`, in an isolated SEARCH_PROBE world, driven through the app's own bench socket. An extension whose manifest declares only `storage` ran:

```js
chrome.runtime.sendNativeMessage("search", {api:"history.search", args:[{text:"", maxResults:5}]})
```

and got real history entries back, titles and URLs included. The same shape answers `tabs.describe` (every open tab's URL and title), `bookmarks.getTree`, `downloads.search` and `sessions.getRecentlyClosed`, none of it mentioned in the install prompt.

## The change

One gate in `ExtensionShims.run`, before the dispatch switch, on the families whose answers leave the extension's own origin:

- `gates` maps the seven families to the Chrome permission of the same name, plus one exact-name entry: `tabs.describe` is the single call where the permission guards reading a tab rather than moving or selecting it, so `tabs.move`/`discard`/`activate` are left alone.
- `allowed(id)` reads the names the extension's own manifest asked for, plus any optional permission already granted through `permissions.request` (the same `extensions.granted.<id>` list the shim's own `granted()` consults), so an extension that asks at runtime is not broken by this.
- A family that was never asked for is refused the way Chrome refuses an API it lacks, so an extension's own error paths see it.

The checks inside the shim stay as they are; they read well as UX. This makes the native side the boundary, which is where a boundary can actually hold.

## Testing

- `swift build` clean, no warnings added.
- In a test world: the no-permission extension above now gets `{"error":"The extension never asked for “history”"}`, and a control extension that declares `history` in its manifest still receives its answer, unchanged.
- Extension code that declares what it uses sees no difference: the store extensions this layer exists for (password managers, Raindrop) declare their permissions.

If you would rather keep the script as the only gate on purpose, that is your call to make and I will close this; from the outside this reads as one data path with a check a hostile caller can walk around.

---


driceroland:
Thank you @Hyp4tia, real hole, well spotted. Two tweaks: the permissions are read from `context.webExtension.manifest` (already in memory) rather than re-reading manifest.json on every call, and `tabs` honours a permission granted later through `permissions.request` (`context.hasPermission(.tabs)`). Tested with three extensions (none, declared, optional-then-granted). Merging!

Hyp4tia:
Re-read the merged version and your tweak is the better shape. Reading `context.webExtension.manifest` keeps the answer in memory instead of a file read on every call, and using `context.hasPermission(.tabs)` for `tabs.describe` is the only one of the two that reflects what the person actually agreed to, optional grants included. My version would have missed a grant made later through `permissions.request`.

One seam worth your judgement, one step further on: for the seven families the gate still answers "did the extension ask", which today is the same as "was it granted" because `load()` grants what the manifest requests. When #135 (a folder reload growing permissions without asking) gets a home, the persisted grant list becomes the record of consent; if the gate reads that and treats the manifest as a ceiling, it stays consent-true on the same line. Not a hole you have now, just where I would look next.

Two notes from my side. The merged tree compiles for me at `2b544f9` (`swift build`, exit 0; the only warning is the pre-existing unused `fcntl` result in Bench, not from any of these five). And the other reports from that pass are #133 to #136, each with a suggested direction (local TLS prompts, the vault's host matcher, the reload consent above, the appcast's DMG link). Yours to take or drop, no rush on them.

PR #129, opened 2026-09-24
Extension packages: refuse links, and never write through one

Hyp4tia:
## What

`Crx.unpack` runs `ditto -x -k` and checks only that `manifest.json` exists (`Crx.swift:117-135`). `ditto` keeps symbolic links as links (verified on this Mac), and `ExtensionShims.prepare` then walks the tree and rewrites every `.html`/`.htm` it finds: `String(contentsOf:)` follows a link, and `write(to:options:.atomic)` replaces the link at the destination with a regular file carrying the target's bytes (`ExtensionShims.swift:118-131`). The service worker path has the same read-and-rewrite shape with the path named by the manifest (`ExtensionShims.swift:75-91`).

Demonstrated on `bbc0c6e`: a folder extension shipping `evil.html` as a symbolic link to a file outside the package. After install, `Extensions/<id>/evil.html` was a regular 72-byte file whose contents were the link target's, with the shim's script tag prepended. The extension can fetch `chrome-extension://<id>/evil.html` and send it anywhere, so the install step turns a signed package into a local file read. The signed CRX makes the entry name attacker-chosen and the same shape crosses `unpack`, so a store-installed package does it too.

## The change

- `Crx.unpack`: after `ditto`, walk the unpacked tree and refuse the package whole if anything in it is a symbolic link. Nothing an extension needs is one, and refusing at the boundary keeps the rest of the code honest.
- `ExtensionShims.prepare`: skip links in the page walker, and resolve the service-worker path against the package folder (`inside(_:of:)`), leaving a manifest path that points outside its own package alone.

Folder installs are copied in as they are, so the walker skip covers those; the unpack refusal covers the store path.

## Testing

- `swift build` clean.
- Test world, same folder extension as above: `main` turns `evil.html` into the copied file, the fixed build leaves it a symbolic link and never reads the target.
- A zip carrying a symbolic link is verified to keep it a link through `ditto -x -k`, which is why the unpack check exists at all.
- Installed new folder extensions before and after: both load and run exactly as before.

---


driceroland:
Thank you @Hyp4tia! Added one more case: a folder install keeps links, so a service worker file that is itself a link (or reached through a linked folder) is skipped too. Merging!

Hyp4tia:
Your extra case is the one I missed: a service worker that is itself a link, and a folder on the way that is one. The resolution test closes the second half neatly, since a linked folder resolves outside the package even when the path itself does not.

Both doors are shut in the merged tree, for the record: the page pass still skips links while it rewrites HTML, and `Crx.unpack` still refuses a package carrying one whole, so a store package cannot arrive with a link inside it. Folder installs are covered by the two guards in the shim, which is where they mattered.

For completeness, since I looked: hardlinks are not a way round it. A zip cannot carry one, and a hardlink inside a folder install is somebody's own arrangement of their own files, not a package reading something it was not given.

PR #130, opened 2026-09-24
Bench: keep a script to the tabs it opened

Hyp4tia:
## What

`Bench.find` matches any tab by id prefix (`Bench.swift:1177-1180`), so with Settings › General › Let a script drive Search on, `eval`, `text`, `click`, `type`, `submit` and `shot` will act in a tab somebody is using: read an authenticated page, or run JavaScript in it. Only `close` checks `tab.bench` (`:265`), and `tap`/`press`/`select` are test-run-only, so the reading and clicking verbs are the reachable ones in a real session.

The socket itself is careful (0600, same-uid `getpeereid`, off until asked for), so what this is about is another process running as the same user: a postinstall script, a build tool, anything that can open a unix socket. Your own README says the bench is what tests this browser "while somebody is using it", and answering inside that person's tabs is the part that does not fit that sentence.

## The change

One line: the lookup matches bench tabs only, which is what the flask mark is for. `close all` and the tab list already draw that line; now the rest agree.

## Testing

- `swift build` clean.
- A test world driven through `./bench` with the change: bench tabs behave exactly as before (open, go, wait, text, eval, click, type, shot, sleep, close), and an id belonging to a tab the bench did not open is answered as an unknown tab (`no tab “…” — see tabs`) instead of being acted on.

If some of your own runs read a real tab on purpose and you would rather keep that, tell me and I will close this; the alternative is to gate the reading verbs on a `--test` run, the way `tap` and `select` already are.

---


driceroland:
Thank you @Hyp4tia! One tweak: the restriction applies outside test runs; in a `--test`/named test world the bench can still reach tabs opened from its own (OAuth popups, `window.open`), which our tests rely on. Merging!

Hyp4tia:
Agreed, and your scoping is the right one: whenever `Store.testing` is true the run is in a world of its own ("Search (test)" or a named one), so "every tab" cannot mean anybody's tab, and the popups a test opens with `window.open` carry no flask, which my version would have locked out of `text`/`eval`/`shot`. Outside test runs the flask rule stands, which is the case that mattered.

The rest of the surface is unchanged, so the real world stays narrow: `tap`, `select`, `press` and `key` are still test-run-only verbs, and `close` keeps its own flask check.

PR #131, opened 2026-09-24
Image menu: only the schemes a picture arrives by

Hyp4tia:
## What

The picture menu's relay hands the page's own `currentSrc` to three native sinks after only `URL(string:)` (`ImageMenu.swift:41`, `:52`):

- open in a new tab (`:70`), where nothing validated the scheme;
- `URLSession.shared.data(from:)` for Copy Image (`:98`);
- `webView.startDownload` (`:114`).

A page names that address, so a `file:` URL reaches all three: a right-click on a crafted `` can pull a local file through the app's URLSession (and onto the pasteboard as a decoded image), or ask WebKit to download it. Nothing else the menu offers is more than a picture, and a picture arrives by `http`, `https` or `data`.

## The change

One clause in the relay's guard: those three schemes pass, anything else gets no menu. Copy Image Address keeps working for whatever a page names.

## Testing

- `swift build` clean.
- The guard is the whole change and it is the same allow-list shape the rest of the app already uses for page-named addresses (`Address.ours`, the favicon fetcher's scheme check at `Icons.swift:195-196`).
- Not driven through the native menu: the menu is modal, and the bench can't right-click yet. If you want a run before merging, say the word and I'll drive one with `hit` and a temporary build that logs the relay's messages.

---


driceroland:
Thank you @Hyp4tia! Two tweaks: `blob:` pictures are allowed too (WhatsApp Web and generated images would otherwise lose their menu entirely), and the page script checks the scheme before `preventDefault()`, so any other scheme keeps WebKit's own menu. The native allow-list stays. Merging!

Hyp4tia:
Both tweaks are better than what I sent. `blob:` is page-generated data with no reach into the file system, so keeping the menu for it costs nothing, and checking the scheme before `preventDefault()` is a nicer answer than mine: an `ftp:`, `file:` or app-scheme picture now falls through to WebKit's own menu instead of getting nothing at all.

The part I care about is unchanged, and it is the one that was the finding: the native allow-list is the gate, since any page can post to `officeImages` directly, and `file:`, `smb:` and app schemes are refused there whatever the page script did.

PR #132, opened 2026-09-24
OAuth: only the tab that began the flow may finish it

Hyp4tia:
## What

`ExtensionAuth.intercept` takes any navigation whose host ends `.chromiumapp.org`, and if a flow for that extension id is waiting, hands the URL to the extension as the sign-in's result (`ExtensionShims.swift:2632`). It never checks that the navigation happened in the tab the flow opened: `entry.tab` is only used to close whatever tab the redirect landed in (`:2637`).

Any page a person visits while a flow is pending can navigate to `https://<id>.chromiumapp.org/?code=…` and have its URL delivered as the flow's answer. The extension then acts on the attacker's code, which is the confused-deputy shape you do not want in the one place a browser handles sign-ins.

## The change

`intercept` takes the navigating web view, and only the flow's own tab may finish it. The call site is `decidePolicyFor`, where `webView` is the view that is navigating by definition, and `Browser.tab(for:)` already answers this question elsewhere.

## Testing

- `swift build` clean.
- Not driven end to end (a real `launchWebAuthFlow` needs a live provider); the guard is the change, and a redirect that arrives in any other tab is left alone rather than consumed, so nothing else changes about how it loads (nothing ever loads from `chromiumapp.org`).

---


driceroland:
Thank you @Hyp4tia! One tweak so providers that open a popup still work: a tab whose opener is the flow's tab may finish it too (and the popup closes with it). Tested with a real `launchWebAuthFlow`. Merging!

Hyp4tia:
The popup case is right, and it does not widen the boundary: what can finish a flow is still "a page running in the tab the flow began in", or a window that page opened. A page in that tab could already finish the flow by navigating itself, since that is what a completed sign-in looks like, so a `window.open` to the redirect adds no reach. Nested popups, where the opener is the popup rather than the flow's tab, stay refused.

Thank you for closing the popup with the flow's tab as well. Left behind, it would have been a tab holding a redirect that never loads, which is a confusing thing to hand somebody.

The assumption the fix rests on, written down for whoever reads it next: the flow's tab is trusted, and what runs in it is the provider. A hostile page that somehow lands in that tab is inside the boundary by definition, and that was true before this change too.

PR #140, opened 2026-09-24
Claim XHTML pages so Search can be the default browser

### LINKED ISSUE #139: Search never appears as a choice for the default web browser
## What I did

Installed Search 1.0.1 (build 202609232336) from the disk image into `/Applications`, opened it, then opened **System Settings → Desktop & Dock → Default web browser**.

## What I expected

Search in that menu, next to the other browsers installed on this Mac.

## What happened instead

The menu lists Brave, Chromium, two copies of Google Chrome for Testing, and Safari. Search is not there.

Launch Services has registered the app. It is a notarized Developer ID build (`com.officecommun.search`), it is flagged as a web browser, and it is a handler for `http`, `https`, and `public.html`. It is not a handler for `public.xhtml`. Every browser that does show up in the menu claims `public.xhtml` as well as `public.html`. Since macOS 11, that second claim is what gets an app into this menu; `http` and `https` alone are not enough.

The comment in `build.sh` says owning `http` and `https` is what lets macOS offer the app as the default browser. The plist that comment sits on only declares `public.html` and `com.apple.web-internet-location`.

## macOS

macOS 26.6.1 (25G76)

Aaditya2605:
## Summary

Fixes #139.

macOS only puts an app in **System Settings → Desktop & Dock → Default web browser** when the app claims `public.xhtml` as well as `public.html`. Search 1.0.1 already claims `http`, `https`, and `public.html`, and Launch Services registers it as a web browser, but it never claims `public.xhtml`, so it is left out of that menu. Brave, Safari, Chromium, and Chrome for Testing on the same Mac all claim `public.xhtml` and all appear.

`build.sh` now declares an XHTML page document type next to the HTML one, and the comment that said owning `http` and `https` was enough to be offered as the default browser now says the XHTML claim is required too.

## Test plan

- [ ] `./build.sh` produces a `Search.app` whose `Info.plist` lists `public.xhtml` under `CFBundleDocumentTypes`
- [ ] Copy that app to `/Applications`, eject any Search disk image still mounted, and run `lsregister -f /Applications/Search.app`
- [ ] Quit and reopen System Settings. **Desktop & Dock → Default web browser** lists Search
- [ ] Choosing Search makes a link from another app open in Search


driceroland:
Merged, thank you! Checked together with today's other PRs in a combined build, in a hidden test world: the built Info.plist carries public.xhtml beside public.html. Opening a local .html/.xhtml from the Finder is a follow-up (Links only takes http/https).

PR #144, opened 2026-09-24
Leave ⌘⇧V to a text field, and open Paste and Go in a new tab

### LINKED ISSUE #138: Cmd+Shift+V copies into a new tab rather than onto a doc
Very simple bug, I was on a google doc trying to CMD+SHIFT+V to paste something, it kept opening a new tab and pasting it in.

Aaditya2605:
## Summary

Fixes #138.

⌘⇧V is paste without formatting in a text field. Brave does not bind that key, so a Google Doc receives it and pastes. Search was taking the key first and loading the clipboard in the tab already open.

- While the caret is in a text field — a form, the address field, or a document that types into its own frame, such as Google Docs — Search leaves the key alone.
- Otherwise it is still Paste and Go, and the result opens in a new tab beside the current one.

Control+Command+V is not the paste shortcut. Ordinary paste is ⌘V. Paste without formatting, in Google Docs and in most Mac apps, is ⌘⇧V. Brave’s “Paste and Match Style” is ⌥⇧⌘V.

## Test plan

- [ ] On a Google Doc, with the caret in the document, ⌘⇧V pastes the clipboard without formatting and does not navigate
- [ ] In a form field, ⌘⇧V is not Paste and Go
- [ ] On a normal page, with nothing being typed, ⌘⇧V opens the clipboard address or search in a new tab and leaves the current tab alone
- [ ] Tabs → Paste and Go does the same


SamarthaB10:
thanks !

SamarthaB10:
Posting on Samartha's behalf; an AI agent wrote this comment for me.

Thanks for this focused fix. It looks like it addresses the specific case in #138. I opened #147 to track the wider shortcut audit, so this PR can stay focused on `⌘⇧V`.

Could you also confirm that a focused editor receives `⌘⇧V` with the menu key equivalent still installed? When the event monitor returns the key for an editor, the `Paste and Go` menu item still has `⌘⇧V`. I am not sure whether AppKit will then activate the menu item before the web view receives the key. The Google Docs check in the test plan should cover this path.

Aaditya2605:
@SamarthaB10 Changes have been made to accomodate multi-use of ⌘⇧V. It now works as a paste-without-formatting command in text boxes and as a paste-and-go command elsewhere. Please have a look and let me know if it works for you.

driceroland:
Thank you @Aaditya2605, the `inputContext` detection is clever. I kept the text-field fix and took out the change that made Paste and Go open a new tab: from an empty tab it left the empty tab behind, and from a private tab it opened an ordinary one. Paste and Go stays in the current tab. Merging!

PR #146, opened 2026-09-24
Fix Proton Pass autofill and conditional passkey detection

ChloeVPin:
## Summary

- Add a timer fallback for `requestIdleCallback` so Proton Pass can finish form detection in WebKit.
- Keep the first credentials container and report conditional passkey support after another handler installs its own `get`.

## Changes

- `ExtensionShims.swift` adds cancellable idle callbacks and preserves `navigator.credentials` before the Chrome shim guard.
- `Passkeys.swift` checks for another own `get` handler when sites ask about conditional mediation or client capabilities.
- `CHANGELOG.md` records both fixes.

## Platforms

- macOS 14 and later: affects WebKit pages in Search.
- Windows and Linux: Search does not support these platforms.

## Test Plan

- `swift build` passed.
- Ran `./build.sh debug app` and completed a passkey sign-in through Proton Pass on Google sign-in.
- Checked the connected Search build on X, GitHub, Google, and Microsoft. Proton Pass showed matching saved logins on each, including in GitHub's password field; Dropbox and Reddit had no matching saved login.


driceroland:
Thank you @ChloeVPin — great catch on `requestIdleCallback`: Proton Pass's form detector calls it directly, WebKit has none, so detection died before finding a field. Confirmed here (`typeof requestIdleCallback` is `undefined` in Search's pages, and the call sits right in Proton Pass's detector).

One change before merging: while this was open, main got the same conditional-passkey answer in `Passkeys.swift` (48fcc2b: `isConditionalMediationAvailable` and `getClientCapabilities` say yes once an extension has claimed `navigator.credentials`, with an equivalent check). The two conflicted, so I kept main's version there and took your `ExtensionShims.swift` part as is: the idle-callback fallback and keeping the first credentials container. I merged main into your branch for that and folded the two changelog lines into one, crediting you.

Builds; Proton Pass installs and loads cleanly with it. I couldn't check the filling itself without a Proton account, so thanks for testing that on Google. Merging!

PR #148, opened 2026-09-24
fix(keyboard): let page editors keep editing shortcuts

### LINKED ISSUE #147: Preserve page editing shortcuts while keeping Search browser shortcuts
## Why

Issue #138 reports one example: `⌘⇧V` in Google Docs starts Search's Paste and Go action instead of pasting in the document. PR #144 adds a focus check for Paste and Go. This audit found other Search shortcuts that can take keys from a page editor or a normal Chrome page. The problem is in Search's app-wide key handling. It is not specific to Google Docs.

Search uses a local key event monitor for the whole window. It handles browser shortcuts before the page receives the key. The handler does not check which page control has focus. The menu also defines Paste and Go with `⌘⇧V`, so test that route as well as the event monitor when reviewing #144.

## Examples

| Keys | Search action | Expected page or Chrome action |
|---|---|---|
| `⌘⇧V` | Paste and Go | Google Docs and other editors use Paste without formatting |
| `⌘⇧←` / `⌘⇧→` | The arrow key code runs Back or Forward | Extend a text selection |
| `⌘S` | Fold the sidebar when it is shown | Save the page or document |
| `⌘D` | Duplicate the tab | Bookmark the page |
| `⌘⇧R` | Reading Mode | Reload without cache |
| `⌘⇧B` | Bookmark the current page | Show or hide the bookmarks bar |
| `⌘⇧H` | Hide page elements | Open the home page |

Google lists its built-in macOS shortcuts in [Chrome Help](https://support.google.com/chrome/answer/157179?hl=en-HK). Google Docs lists its editor shortcuts in [Docs Help](https://support.google.com/docs/answer/179738?hl=en). Websites and extensions can add more shortcuts.

## What I did

1. Opened a Google Doc and put the caret in the document.
2. Pressed `⌘⇧V` to paste without formatting.
3. On a page text field or editor, pressed `⌘⇧←` or `⌘⇧→` to extend the selection.
4. Used common page commands such as `⌘S` or `⌘D` while page content had focus.

## What I expected

The focused page editor should receive its editing shortcuts. Search should keep its tab, history, and other browser shortcuts when those keys do not conflict with page editing. Each overlapping key should have a clear rule. A fix for one key should not disable the other Search shortcuts.

## What happened

The current handler takes `⌘⇧V` for Paste and Go without checking page focus. Its Command-arrow fallback checks the left and right arrow key codes but does not reject Shift, so it can take text selection as Back or Forward. Other browser bindings also overlap with common Chrome or document commands.

There is a mismatch to check for `⌘⇧V`: issue #138 says it opens a new tab, but the current `pasteAndGo()` code loads the destination in the active tab. PR #144 describes the old behavior the same way, then changes Paste and Go to open a new tab when the page is not being edited. Please reproduce the reported behavior before treating the new-tab result as confirmed.

## Work requested

Audit Search's current key bindings against Chrome's macOS shortcut list and common macOS and web editor commands. Route each key to the focused page or to Search by a consistent rule. Keep useful Search bindings when page text is not being edited. Make modifier checks exact. Keep the menu shortcuts and the event monitor consistent.

Check #144 for the `⌘⇧V` fix. In particular, verify that a focused editor still receives the key when the Paste and Go menu item has the same key equivalent. Keep the separate Paste and Go command available when its browser shortcut is safe to use.

## Done when

- `⌘⇧V` pastes in Google Docs, a normal text field, and a content-editable editor.
- `⌘⇧←` and `⌘⇧→` extend text selection. Unshifted `⌘←` and `⌘→` still navigate browser history.
- The listed conflicting keys have an intentional and documented owner in page-editing and normal page states.
- Existing Search tab and browser shortcuts still work when the page is not editing text.
- The check covers a normal page, text field, content-editable editor, Google Docs, and a page with no editor focused.

macOS version: 26.6.2.

Related: #138, #144.

SamarthaB10:
## Problem

Search takes some shortcuts before a focused page editor can use them. A menu key equivalent can also take a key after the local monitor passes it to WebKit.

## Change

- Give focused page editors and Search text fields `⌘⇧V`, `⌘S`, `⌘D`, `⌘⇧R`, `⌘⇧B`, and `⌘⇧H`. Search keeps those commands when no editor has focus.
- Relay editor focus through the active iframe, including frames from another origin. Reports from inactive frames do not change shortcut routing.
- Remove the six matching menu key equivalents so a passed key reaches the editor. The menu commands remain available by click.
- Document the focus rule in the README and add an Unreleased changelog entry. The current main branch already rejects Shift in the `⌘←/→` history fallback.

Fixes #147. Related: #144.

## Verification

- `swift build` passed. It reports warnings in unchanged files; the changed files add none.
- `git diff --check origin/main...HEAD` passed.
- Both injected JavaScript blocks passed a Node syntax check.
- A Search test world key check was inconclusive because other keyboard activity reached the test window. Google Docs and live page-editor behavior still need a manual check. This package has no SwiftPM test target.

Model: GPT-6 Codex. Harness: Codex CLI, SwiftPM, Node, and Search bench test world `keyboard-147`.


driceroland:
Thank you @SamarthaB10! #144 fixes ⌘⇧V in text fields (it actually pastes, where handing the key back to WebKit pastes nothing), and removing six shortcuts from the menus goes further than we'd like. So I'm closing this one. Letting the page handle keys first, as Chrome does, would make a good issue to discuss on its own.

PR #149, opened 2026-09-24
The cursor returns over a panel above a page that hides it

0froq:
`cursor: none` on a page leaked through every panel that floats on top of it. Open a page that hides its cursor (some video and game sites; anything with `body { cursor: none }`), press ⌘, — Settings came up with the pointer invisible across the whole dimmed backdrop and the panel itself, until you stepped off. Same on History, Downloads, Bookmarks, Welcome and Passwords, since they share `sheet` in `App.swift`.

The cause is how AppKit resolves a cursor. WebKit answers CSS `cursor: none` with an `NSCursor.invisibleCursor` rect over the whole web view; the SwiftUI panels in `sheet`'s ZStack add no rect of their own, so the invisible one is still the frontmost answer wherever the pointer sits inside the sheet. `CursorGround` is a small `NSViewRepresentable` at the foot of every sheet's ZStack that adds an arrow cursor rect across its bounds, and re-arms on `viewDidMoveToWindow` and `layout` so AppKit recomputes the moment the sheet appears rather than after the pointer twitches. One place fixes all six, and no panel has to know it was the fix.

`Float.swift` already uses `resetCursorRects` for its corner handle, so the shape is the app's own.

Checked on macOS 26.5 against a local `cursor: none` page: the arrow is back across all six panels the moment the pointer crosses into the sheet, without moving first. `swift build` adds no warnings.


driceroland:
Thank you @0froq, neat fix in one place. I couldn't watch the cursor in a hidden test run, but WebKit's source leaves the cursor alone while an AppKit cursor rect is in effect, so the arrow should hold while moving over a panel too. Merging!

0froq:
Glad it helped. Thank you for such a well-made little browser — happy to add a line to it.

Issue #123, opened 2026-09-24
PIP and extension


jercydy29:
Vimnuim extension is not working properly
PIP keep on showing up at the bottom right even after moving it somewhere else it annoying please fix

PR #150, opened 2026-09-24
PiP: the floating video window opens where it was last left

farchanrifai:
Fixes the placement part of #123: "PIP keep on showing up at the bottom right even after moving it somewhere else".

## What it does

- The floating window's frame is kept as it moves and resizes, and again when it closes, in `Store.settings` under `float.frame`.
- When a video floats, the window opens at that place and size, across closing it and quitting.
- If no screen still shows most of that frame (a display unplugged, say), it starts in the bottom-right corner as before.
- The aspect ratio and minimum size are unchanged.

A fix, so no switch. CHANGELOG gets its line under Unreleased › Fixed.

## Tested

In my fork, by hand: moved and resized the window, closed it and floated again; it came back where and as big as it was. The off-screen fallback was not tried by hand.


driceroland:
Thank you @farchanrifai! One change: the frame is saved at the end of a resize, when the window closes and when the app quits, instead of on every move step (with the flick animation that was ~120 writes a second). Merging!

PR #151, opened 2026-09-24
PiP: two fingers can flick the floating video to a corner

farchanrifai:
As in Arc and Dia: a two-finger swipe on the floating window sends it to the corner or edge it points at.

## What it does

- **Straight swipes:** up takes the window to the top on the side it's on; left takes it to the left at the height it's at, and so on.
- **Diagonal swipes** (between about 22° and 68°) go to that corner.
- **Placement:** the window lands 12 points in from the edges of the screen's usable part.
- **One move per swipe,** however long. The swipe is read as a whole when the fingers lift, so one that starts straight and turns diagonal still reaches its corner. A long swipe (over 120 points) moves without waiting.
- **Mouse wheel:** one flick per notch.
- **The glide** runs off the display's own refresh (`CADisplayLink`, 120 Hz on ProMotion) on a critically damped spring: quick away, no overshoot. A click stops it.
- **Dragging** still puts the window anywhere.

The rule is one pure function, `Float.corner(for:in:toward:margin:)`.

## Switch or default? Your call

Following CONTRIBUTING, it's behind **Settings › General › Flick the floating video to a corner**, off by default. With it off, two fingers push the window along as today.

It might suit Search better as the only behaviour, since the gesture only exists on the floating window. If you'd rather, I'll drop the switch and the old push-along path. Happy either way.

CHANGELOG gets its line under Unreleased › Added.

## Tested

- By hand in my fork (where it's on without a switch): straight and diagonal flicks on a trackpad. The mouse wheel and a click mid-glide weren't tried by hand.
- On this branch: it builds. The corner rule has unit tests in my fork; this repo has no test target, so they're not included.


driceroland:
Merged, thank you! Checked together with today's other PRs in a combined build, in a hidden test world: it builds, and the switch is off by default in Settings › General.

PR #152, opened 2026-09-24
PiP: the video can float when you switch apps

farchanrifai:
As in Arc (Dia doesn't do it): switch to another app while a video plays, and it comes out into the floating window. It goes back into its tab when you return.

## What it does

- **Switching away:** when another app comes to the front, the video on the tab you're on floats out, the same way it does today when you leave the tab. It only happens on the same sites (`Players.knows`) and only while something plays.
- **Coming back:** if you're still on the tab the video came from, it lands back in the tab.
- **⌘⇧P:** a video floated by hand stays out.
- **Several windows:** only the window last in front lifts its video, so you don't get one floating window per browser window.
  - That window is tracked by hand (`Browser.front`, set on `didBecomeKey`). Once the app has gone to the back, `isMainWindow` no longer answers, which I found out the hard way.

## Switch or default? Your call

Following CONTRIBUTING, it's behind **Settings › General › Float the video when you switch apps**, off by default. It's the same rule the browser already applies when you leave a tab, so it could also just be how floating works. If you'd rather have it that way, I'll drop the switch.

CHANGELOG gets its line under Unreleased › Added.

## Tested

- By hand in my fork (on without a switch): a video playing, then switching to another app (it floats) and back (it lands). The ⌘⇧P case wasn't tried by hand.
- On this branch: it builds. The Settings switch wasn't tried by hand.


driceroland:
Merged, thank you! Checked together with today's other PRs in a combined build, in a hidden test world: it builds beside #151's setting, both off by default; a video lifted with ⌘⇧P stays out when you come back (appLeft leaves it alone, checked in the code). The float itself shows on screen, so trying it by hand is for Drice.

PR #154, opened 2026-09-24
Keep different articles and videos in History

yuxino:
Visit two Hacker News articles, such as `item?id=8863` and `item?id=1`, then open History. Only the second article remains: the first was overwritten because both addresses were shortened to `/item` before being stored. The same happens with different YouTube video IDs.

Before (real Search app):

https://github.com/user-attachments/assets/f92af301-f8bc-49b2-bd25-6c2b2640121f

After:

https://github.com/user-attachments/assets/f3078469-c16a-4e33-afbf-2a3fc117b2d1

History now keeps the query parameters, including their case, when identifying a page. Existing entries are read using their saved URLs.

Checked in the macOS app: both articles stay in History and the earlier one opens correctly. Also checked revisits, title updates, removal, imports and loading existing history. `swift build` passes.


driceroland:
Landed on main as 19ab9c9, with you as the author. Thank you @yuxino! (Your branch wasn't open to maintainer pushes, so it went in directly rather than through the merge button.) Tested in a hidden test build: two Hacker News-style items (?id=8863, ?id=1) and two video ids differing only in case (?v=AbC, ?v=abc) stay four separate entries, and a repeat visit counts on its own entry. One addition on top: when an older history file loads, two entries that end up with the same new key are now merged (counts added, the most recent kept) instead of going through Dictionary(uniqueKeysWithValues:), which would have stopped Search at launch. It's in the changelog for the next update.

PR #155, opened 2026-09-24
look for other browsers only when the welcome screen needs them

dttdrv:
the first launch walked every Chromium browser's folders before its window could appear, because `WelcomePanel` used `Chromium.installed().first` as the initial value of `source`. SwiftUI evaluates that whenever the view is made, so it ran while the first window was being built, for a page that isn't showing yet. `source` now starts as nil and Bring them in falls back to the first installed browser, like the picker already does.

first launch in a fresh test world went from 313 ms to 266 ms, median of 10 interleaved hidden runs on an M5 Pro. later launches are unchanged since they don't build the panel.


driceroland:
Merged, thank you! Checked together with today's other PRs in a combined build, in a hidden test world: it builds and launches, first launch included.

dttdrv:
awesome! looking forward to using this instead of helium!

PR #156, opened 2026-09-24
find each profile's Login Data without walking the caches

dttdrv:
`Chromium.Source.files` walked three levels into every profile, caches included, to find `Login Data`. Chromium keeps it directly inside each profile folder, so this lists the root and checks each profile instead. it finds the same files.

`Chromium.installed()` runs this for every known browser, and Passwords, Bookmarks and the welcome screen call it each time they're drawn. with Dia installed it went from about 18 ms to 3 ms per call on an M5 Pro, and from 87 ms to 3 ms cold.


driceroland:
Merged, thank you! Checked together with today's other PRs in a combined build, in a hidden test world: it builds and launches; the import lists read as before.

